PR #3, opened 2017-04-19
Parsing

clenk:
Parse all SDOs and SROs.

What is not parsed (yet):
- Cyber Observable Objects
- Objects without a 'type' property, eg. External References by themselves, that aren't part of an SDO


stix2/properties.py
```diff
@@ -106,23 +139,72 @@ def __init__(self, type):
...
+
+class TimestampProperty(Property):
+
+    def clean(self, value):
+        if isinstance(value, dt.date):
+            if hasattr(value, 'hour'):
+                return value
+            else:
+                # Add a time component
+                return dt.datetime.combine(value, dt.time(), tzinfo=pytz.timezone('US/Eastern'))
```
gtback: Also, pytz timezones [shouldn't be passed](http://pytz.sourceforge.net/#localized-times-and-date-arithmetic) to `tzinfo` of datetime constructors: 

> Unfortunately using the tzinfo argument of the standard datetime constructors ‘’does not work’’ with pytz for many timezones.

stix2/other.py
```diff
@@ -0,0 +1,124 @@
...
+        if type(value) in [TLPMarking, StatementMarking]:
+            return value
+        else:
+            raise ValueError("must be a Statement or TLP Marking.")
+
+
+class MarkingDefinition(_STIXBase):
+    _type = 'marking-definition'
+    _properties = {
+        'created': TimestampProperty(default=lambda: NOW, required=True),
```
gtback: `required` and `default` shouldn't be used together. `required` really means "the user must provide this" (and we can't or don't want to create a default value). `default` implies that the field is required in the spec. 

I know this causes confusion, so we might want to consider renaming or documenting this better.

stix2/properties.py
```diff
@@ -17,82 +23,109 @@ class Property(object):
...
-    def __init__(self, contained, required=False, element_type=None):
+    def __init__(self, contained, **kwargs):
         """
-        contained should be a type whose constructor creates an object from the value
+        Contained should be a function which returns an object from the value.
         """
-        self.contained = contained
-        self.element_type = element_type
-        super(ListProperty, self).__init__(required)
+        if contained == StringProperty:
```
gtback: Could we just define `__call__` on `StringProperty` and `BooleanProperty` to avoid this chain of `if`/`else`?

stix2/properties.py
```diff
@@ -1,5 +1,11 @@
 import re
 import uuid
+from six import text_type
```
gtback: Can we split these into built-in (standard library) imports, third-party imports (six, dateutil), and internal (`.base`), and alphabetize within each group? I think there are ways to check for this in CI.

stix2/__init__.py
```diff
@@ -3,11 +3,48 @@
...
+from .utils import get_dict
 from . import exceptions
+
+
+def parse(data):
+    """Deserialize a string or file-like object into a STIX object"""
+
+    obj = get_dict(data)
+
+    obj_map = {
```
gtback: We can pull this out of the function, as a (minor) performance optimization.

stix2/other.py
```diff
@@ -0,0 +1,124 @@
...
+class TLPMarking(_STIXBase):
+    # TODO: don't allow the creation of any other TLPMarkings than the ones below
+    _properties = {
+        'tlp': Property(required=True)
+    }
+
+
+class StatementMarking(_STIXBase):
+    _properties = {
+        'statement': Property(required=True)
```
gtback: Should this be a StringProperty?

stix2/properties.py
```diff
@@ -17,82 +23,109 @@ class Property(object):
...
         - provide a default value for this property.
         - `default()` can return the special value `NOW` to use the current
             time. This is useful when several timestamps in the same object need
             to use the same default value, so calling now() for each field--
             likely several microseconds apart-- does not work.
 
-    Subclasses can instead provide lambda functions for `clean`, and `default`
-    as keyword arguments. `validate` should not be provided as a lambda since
-    lambdas cannot raise their own exceptions.
+    Subclasses can instead provide a lambda function for `default as a keyword
```
gtback: Missing a "\`"  after `default`.

stix2/test/test_external_reference.py
```diff
@@ -34,7 +35,7 @@ def test_external_reference_capec():
     )
 
     assert str(ref) == CAPEC
-    assert repr(ref) == "ExternalReference(external_id='CAPEC-550', source_name='capec')"
+    assert re.match("ExternalReference\(external_id=u?'CAPEC-550', source_name=u?'capec'\)", repr(ref))
```
gtback: We don't really need these repr tests. I was just trying to be fancy. It also fails pretty quickly on datetimes.

stix2/properties.py
```diff
@@ -17,82 +23,109 @@ class Property(object):
...
-        try:
-            value = self.clean(value)
-        except NotImplementedError:
-            pass
         return value
 
+    def __call__(self, value=None):
+        if value is not None:
+            return value
+
```
chisholm: All code paths in \_\_call\_\_ should return a value, or none should.  If value is None, execution falls off the end without a return (which means None is returned, but that should be explicit).  As written so far, this method is a no-op anyway.  It's the same as the one-liner `return value`.

stix2/properties.py
chisholm: The documentation for self.contained sounds exactly like it's describing Property.clean ("returns an object from the value"... kinda vague, but I think that's what it means).  Why wouldn't that be suitable for all properties?  Also, if contained/clean is responsible for coercing/converting to a particular type, what role does Property.\_\_call\_\_ play?  The method isn't documented, and I can't tell from the code.

Also, creating a throwaway object just so you can call an instance method is ugly.  It implies that if that method ever uses `self`, this code is probably wrong.  But if clean() is not allowed to use `self`, maybe it should be a static or class method.

Finally, can I create a list of lists?  (E.g. ListProperty whose "contained" is another ListProperty.)  The code doesn't check for it, and it looks like it may error out (if contained is ListProperty, contained() is constructing a ListProperty without the correct constructor arguments).  More generally, that code seems to assume that all properties have a no-arg constructor.  And that's not the case.

clenk: In Python, \_\_call\_\_ lets you do things like:

```python
class A():
    ...

a = A() # calls __init__()
a() # calls __call__()
```

This is needed for ListProperty fields. For some, like AttackPattern.kill_chain_phases, ListProperty.contained is set to the KillChainPhase type (and not an instance of KillChainPhase because of KillChainPhase's required properties). But for others, like Sighting.where_sighted_refs, the ListProperty.contained is set to an instance of ReferenceProperty because where_sighted_refs is limited to references to Identity objects and we have to set the ReferenceProperty.type.

clean() could be a static or class method for some Property types, but not all of them. For example, IDProperty.clean() uses self.required_prefix.

You can indeed create lists of lists, but the innermost ListProperty must specify a type. `ListProperty(ListProperty())` will throw a TypeError because the ListProperty constructor requires a parameter, but `ListProperty(ListProperty(StringProperty))` is fine.

stix2/other.py
clenk: I've added a paragraph to Property's docstring about this.

stix2/properties.py
clenk: I thought I fixed that. Well, should be fixed now.

>    It is safe for timezones without daylight saving transitions though, such as UTC

stix2/properties.py
clenk: I've added a flake8 plugin to Tox for this purpose.

PR #4, opened 2017-05-02
Added versioning api, with tests

rpiazza:


stix2/test/test_versioning.py
```diff
@@ -0,0 +1,108 @@
+import pytest
+import stix2
+
+
+EXPECTED = """{
```
gtback: Do we need this anywhere? Can we remove it?

stix2/test/test_versioning.py
```diff
@@ -0,0 +1,108 @@
...
+    "description": "Campaign by Green Group against a series of targets in the financial services sector.",
+    "id": "campaign--8e2e2d2b-17d4-4cbf-938f-98ee46b3cd3f",
+    "modified": "2016-04-06T20:03:00.000Z",
+    "name": "Green Group Attacks Against Finance",
+    "type": "campaign"
+}"""
+
+
+def test_making_new_version():
+    campaign_v1 = stix2.Campaign(
```
gtback: Maybe create a `campaign` fixture in `conftest.py`, then pass that into each of these functions? I know `conftest.py` and pytest fixtures can be strangely magical, so let me know if you run into any problems.

stix2/test/test_versioning.py
```diff
@@ -0,0 +1,108 @@
...
+        created="2016-04-06T20:03:00.000Z",
+        modified="2016-04-06T20:03:00.000Z",
+        name="Green Group Attacks Against Finance",
+        description="Campaign by Green Group against a series of targets in the financial services sector."
+    )
+
+    campaign_v2 = campaign_v1.revoke()
+
+    with pytest.raises(stix2.exceptions.VersioningError) as excinfo:
+        campaign_v3 = campaign_v2.new_version(name="barney")   # noqa
```
gtback: To clean this up a bit, just don't assign the output to anything. Then you can get rid of the `# noqa`. Same for the other tests.

stix2/exceptions.py
```diff
@@ -49,3 +49,10 @@ class ImmutableError(STIXError, ValueError):
 
     def __init__(self):
         super(ImmutableError, self).__init__("Cannot modify properties after creation.")
+
+
+class VersioningError(STIXError, ValueError):
+    """Execption while using the Versioning API"""
+
+    def __init__(self, msg):
```
gtback: Maybe split this into separate exceptions? `AlreadyRevokedError` and `UnmodifiableFieldError`? The latter could take the list of fields as an argument, so we can check that without needing to do exact string matching.

stix2/base.py
```diff
@@ -99,3 +101,33 @@ def __repr__(self):
...
+
+#  Versioning API
+
+    def new_version(self, **kwargs):
+        unchangable_properties = []
+        if self.revoked:
+            raise VersioningError("Cannot create a new version of a revoked object")
+        new_obj_inner = copy.deepcopy(self._inner)
+        properties_to_change = kwargs.keys()
+        if "type" in properties_to_change:
```
gtback: Maybe handle these in a loop?

stix2/base.py
```diff
@@ -99,3 +101,33 @@ def __repr__(self):
         props = [(k, self[k]) for k in sorted(self._properties) if self.get(k)]
         return "{0}({1})".format(self.__class__.__name__,
                                  ", ".join(["{0!s}={1!r}".format(k, v) for k, v in props]))
+
+#  Versioning API
+
+    def new_version(self, **kwargs):
```
gtback: How do we "unset" an attribute? Is passing `None` sufficient? Can we add a test for that?

gtback:
This "request changes" thing is new to me, so we'll see how it works.

stix2/test/test_versioning.py
rpiazza: done

stix2/exceptions.py
rpiazza: Sure - I didn't know how fancy you wanted to get with the exception hierarchy.

stix2/base.py
rpiazza: That was my thought - I will put in a test to make sure it makes sense

stix2/test/test_versioning.py
rpiazza: I assume it won't complain that I'm calling a method that returns a value, but I'm not assigning it to anything????

stix2/exceptions.py
rpiazza: Are you thinking of having intermediary exception classes?  In other words, should those two exception classes you suggest be subclasses of VersioningError?

stix2/base.py
rpiazza: Good suggestion! Done

stix2/exceptions.py
rpiazza: You seem to be using "field", instead of "property" when naming things - is this intended or a slip?

stix2/base.py
rpiazza: ok - that doesn't work - it turns it into the string 'None'.  I think it should be deleted from _inner, but we have to be careful, because it could be a required property

stix2/test/test_versioning.py
gtback: Nope, it shouldn't complain. Right now, I believe it is complaining that you assign to a variable you don't use. It isn't smart enough to tell whether a function has side affects or not, so it has to assume the function "does something" even if you're ignoring the return value.

stix2/exceptions.py
gtback: I want to have exceptions that have meaningful attributes (other than just the string message) where they make sense. For example, the list of properties that the user tried to modify in the new version that aren't allowed. 

I don't think we need a VersioningError layer in a hierarchy, but could be convinced otherwise. 

Yes, I mean "property" every time I say "field".

stix2/base.py
gtback: This is tricky. If a unsets a property that cannot be changed, the function should (and probably already does) raise an Exception. If a user unsets a property that is required with no default implementation (like `Indicator.pattern`), I'm thinking we should let the object initializer raise that error. If a user tries to unset a property that is "required", but that we have a default implementation for (like `Indicator.valid_from`), it might be confusing, since "unset" is acting more like "reset". I'm not sure what the right thing to do is.

stix2/exceptions.py
rpiazza: You named some exceptions using field:  e.g., ExtraFieldError - should be be more consistent with the names of things??

stix2/test/test_versioning.py
rpiazza: Done - and it seems to either be smart or not smart depending how you look at it - because it didn't complain...

stix2/test/test_versioning.py
rpiazza: The fixtures in `conftest.py` seem to be there to test the required fields.  The campaign instance I used was "stolen" from `test_campaign.py` - and you don't use fixtures there that I can see.

stix2/exceptions.py
gtback: Yes, there's a lot more instances of `field` than I realized. Oops. Let's do the right thing (use "Property" here) and I'll make a new issue to go back and clean up the other ones.

stix2/test/test_versioning.py
gtback: The fixtures can be used any time we want to have a "fixed" value that we don't want to create as part of the test itself. It's partially an code-organization thing (don't repeat yourself), and partially because, if we ever need an "example" Campaign for another test anywhere else, we can use it.

rpiazza: Can you show me a place where you are using, for instance, the indicator fixture?

gtback:
@rpiazza: Have you pushed the changes you made to address some of the things above? I think this is about ready to merge at that point.

rpiazza:
I think we should figure out "unset" before we merge this in.

Issue #19, opened 2017-05-31
STIX objects always claim to have all attributes


chisholm:
The reason is that [_STIXBase.\_\_getattr\_\_()](https://github.com/oasis-open/cti-python-stix2/blob/9dbde930314d11f4185d3613412683d0b2a88714/stix2/base.py#L127) returns None if the property doesn't exist.  That's saying that the property exists, with value None.  Instead, it needs to raise AttributeError.

Fixing this will reveal another bug [here](https://github.com/oasis-open/cti-python-stix2/blob/9dbde930314d11f4185d3613412683d0b2a88714/stix2/base.py#L81) about the "granular_markings" attribute not existing.  Attribute existence should be checked with `hasattr()`.

Issue #17, opened 2017-05-30
Can't set a property to False


chisholm:
The culprit seems to be code that is intended to check for None, but is too general.  See [here](https://github.com/oasis-open/cti-python-stix2/blob/33cc4caa3694e18882264d57d64f53b513171d88/stix2/base.py#L100).

That will prevent any property from being settable to False, empty lists, dicts, strings, zero, etc.  It should check specifically for None if that's the intent.

PR #20, opened 2017-06-01
Instantiate ImmutableError, more descriptive message. Check the attribute being requested actually exists in object. Fix being able to set certain empty values like: list, strings, False, etc.

### LINKED ISSUE #13: Forgot to instantiate ImmutableError
I'm guessing [this line](https://github.com/oasis-open/cti-python-stix2/blob/f937e2bb3f104e3cdf0578ed8323b2f7e39119a4/stix2/base.py#L99) should raise an instance of the exception class, not the class itself?

emmanvg:
Related to issue #13, #17 and #19

### DIFF
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -133,7 +133,7 @@ class _STIXBase(collections.Mapping):
             # If attribute not found, check if its a property of the object.
             if name in self._properties:
                 return None
-            
+
             raise AttributeError("'%s' object has no attribute '%s'" %
                                  (self.__class__.__name__, name))
 
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -97,7 +97,7 @@ class _STIXBase(collections.Mapping):
         # Remove any keyword arguments whose value is None
         setting_kwargs = {}
         for prop_name, prop_value in kwargs.items():
-            if prop_value:
+            if prop_value is not None:
                 setting_kwargs[prop_name] = prop_value
 
         # Detect any missing required properties
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -81,10 +81,9 @@ class _STIXBase(collections.Mapping):
             raise DependentPropertiesError(self.__class__, failed_dependency_pairs)
 
     def _check_object_constraints(self):
-        if hasattr(self, "granular_markings") and self.granular_markings:
-            for m in self.granular_markings:
-                # TODO: check selectors
-                pass
+        for m in self.get("granular_markings", []):
+            # TODO: check selectors
+            pass
 
     def __init__(self, **kwargs):
         cls = self.__class__
@@ -127,16 +126,16 @@ class _STIXBase(collections.Mapping):
 
     # Handle attribute access just like key access
     def __getattr__(self, name):
-        # If the requested attribute is not None return its value
-        if self.get(name) is not None:
-            return self.get(name)
-        # Otherwise check the attribute exists inside object. If it exists
-        # return None. If not raise the AttributeError.
-        elif name in self._properties:
-            return None
-
-        raise AttributeError("'%s' object has no attribute '%s'" %
-                             (self.__class__.__name__, name))
+        try:
+            # Return attribute value.
+            return self.__getitem__(name)
+        except KeyError:
+            # If attribute not found, check if its a property of the object.
+            if name in self._properties:
+                return None
+            
+            raise AttributeError("'%s' object has no attribute '%s'" %
+                                 (self.__class__.__name__, name))
 
     def __setattr__(self, name, value):
         if name != '_inner' and not name.startswith("_STIXBase__"):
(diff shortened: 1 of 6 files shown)


emmanvg:
Is there a problem with the tox configuration?

emmanvg:
Figured I should had separated this into different branches and do three different PRs....

clenk:
Looks like there was a bug in isort, which released a new version today, but it just got fixed. Re-running Travis.

stix2/base.py
```diff
@@ -124,11 +124,20 @@ def __len__(self):
 
     # Handle attribute access just like key access
     def __getattr__(self, name):
-        return self.get(name)
+        # If the requested attribute is not None return its value
```
gtback: Would it work to just call `self.__getitem__`, catch the `KeyError`, and raise an `AttributeError` instead?

stix2/exceptions.py
```diff
@@ -47,8 +47,14 @@ def __str__(self):
...
 
-    def __init__(self):
-        super(ImmutableError, self).__init__("Cannot modify properties after creation.")
+    def __init__(self, cls, key):
+        super(ImmutableError, self).__init__()
+        self.cls = cls
+        self.key = key
+
+    def __str__(self):
+        msg = "Cannot modify '{0.key}' property in {0.cls} after creation."
```
gtback: Let's use `cls.__name__` in this message. It will make the messages more human-friendly.

stix2/base.py
chisholm: This code still doesn't achieve consistency between dict and attribute access.  This will say that any *legal* attribute exists (for attribute access), but only *set* keys exist (for dict access).

Greg's suggestion would work (can even just do `self[name]`).  Or just delete the `elif name in self._properties` part.  (Checking `self.get(name)` against `None` technically can't distinguish between the key not existing, and the key existing with value `None`.  But with `None` used for "un-setting" a property, I guess that can't happen.)

emmanvg: That's the problem I was experiencing. Doing `self[name]` or `self.__getitem__(name)` would generate an AttributeError because the key will not exist on the `self._inner` dictionary if there is no value set for that property. So I don't have a clear way to check "is the key there?" and "Is the value of the key not None?". For example a check like `bundle.objects is None` raises an AttributeError if "objects" is not set with the current suggested changes.

stix2/test/test_observed_data.py
```diff
@@ -761,14 +761,13 @@ def test_file_example_with_WindowsPEBinaryExt():
 
 
 def test_file_example_encryption_error():
-    with pytest.raises(stix2.exceptions.DependentPropertiesError) as excinfo:
-        stix2.File(name="qwerty.dll",
-                   is_encrypted=False,
-                   encryption_algorithm="AES128-CBC"
-                   )
+    file_ = stix2.File(name="qwerty.dll", is_encrypted=False,
```
gtback: Why is the result of this test changing? As far as I can tell, it should be invalid to create a File with an `encryption_algorithm` when `is_encrypted=False`. If the changes to `getattr` caused this constraint to stop working, we should fix that.

emmanvg: You are correct. I think I misunderstood what we were testing here. Yes, this is brought up by the changes to `getattr`.

emmanvg: More specifically: https://github.com/emmanvg/cti-python-stix2/blob/4d58fc3dbb7697f951d35a221f2d3dbc8e0e1b5d/stix2/base.py#L100

emmanvg: The code changes in __getattr()__ were not the reason.

stix2/base.py
chisholm: The "is the key there" check could be `if name in self` (or equivalently, `if name in self._inner`).  Using the `in` operator implicitly leverages the dict/key access mechanism.  So the whole solution could be as simple as:
```python
if name in self:
    return self[name]
raise AttributeError(name)
```
Or just attempt the access and catch KeyError, but I prefer to only use the exception mechanism to handle exceptional situations.  (Which admittedly goes against some of Python's fundamental designs, e.g. raising StopIteration when an iterator is exhausted.  Reaching the end of iteration is a completely normal occurrence.)

If dict and attribute access are to behave identically, then `bundle.objects` *must* cause an AttributeError if the `objects` attribute is not set, because that's what dict/key access would do (except with a KeyError).  I've been operating under the assumption that's what you wanted.  That's what the [comment](https://github.com/emmanvg/cti-python-stix2/blob/a38b912d1960245b84c5cb9c9af16fde8e420ac5/stix2/base.py#L125) in the code said.  If so, then if users want to check for property existence via the dict/key mechanism, they can use `if "some_prop" in my_stix_obj`.  If they want to check via attributes, they can use `if hasattr(my_stix_obj, "some_prop")`.  Using `if my_stix_obj.some_prop` as an existence check would be incorrect.

gtback:
This is looking good to me!

PR #25, opened 2017-06-22
Add timestamp precision for `created` and `modified`

### LINKED ISSUE #24: Timestamp format sometimes incorrect for `created` and `modified`
The STIX specification says:

> The created timestamp MUST be precise to the nearest millisecond (exactly three digits after the decimal place in seconds).
> The modified timestamp MUST be precise to the nearest millisecond (exactly three digits after the decimal place in seconds).

This is to avoid ambiguity in whether one timestamp equals another for purposes of versioning.

python-stix2 doesn't seem to always do this (unless I'm missing something).

clenk:
Fix #24

### DIFF
--- a/stix2/properties.py
+++ b/stix2/properties.py
@@ -1,18 +1,15 @@
 import base64
 import binascii
 import collections
-import datetime as dt
 import inspect
 import re
 import uuid
 
-from dateutil import parser
-import pytz
 from six import text_type
 
 from .base import _Observable, _STIXBase
 from .exceptions import DictionaryKeyError
-from .utils import get_dict
+from .utils import get_dict, parse_into_datetime
 
 
 class Property(object):
@@ -215,26 +212,15 @@ class BooleanProperty(Property):
 
 class TimestampProperty(Property):
 
-    def clean(self, value):
-        if isinstance(value, dt.date):
-            if hasattr(value, 'hour'):
-                return value
-            else:
-                # Add a time component
-                return dt.datetime.combine(value, dt.time(), tzinfo=pytz.utc)
+    def __init__(self, precision=None, **kwargs):
+        self.precision = precision
+        super(TimestampProperty, self).__init__(**kwargs)
 
-        # value isn't a date or datetime object so assume it's a string
+    def clean(self, value):
         try:
-            parsed = parser.parse(value)
-        except TypeError:
-            # Unknown format
-            raise ValueError("must be a datetime object, date object, or "
-                             "timestamp string in a recognizable format.")
-        if parsed.tzinfo:
-            return parsed.astimezone(pytz.utc)
-        else:
-            # Doesn't have timezone info in the string; assume UTC
-            return pytz.utc.localize(parsed)
+            return parse_into_datetime(value, self.precision)
+        except ValueError:
+            raise
 
 
 class ObservableProperty(Property):
--- a/stix2/properties.py
+++ b/stix2/properties.py
@@ -217,10 +217,7 @@ class TimestampProperty(Property):
         super(TimestampProperty, self).__init__(**kwargs)
 
     def clean(self, value):
-        try:
-            return parse_into_datetime(value, self.precision)
-        except ValueError:
-            raise
+        return parse_into_datetime(value, self.precision)
 
 
 class ObservableProperty(Property):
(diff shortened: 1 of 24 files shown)


stix2/properties.py
```diff
@@ -215,26 +212,15 @@ def clean(self, value):
...
-            # Unknown format
-            raise ValueError("must be a datetime object, date object, or "
-                             "timestamp string in a recognizable format.")
-        if parsed.tzinfo:
-            return parsed.astimezone(pytz.utc)
-        else:
-            # Doesn't have timezone info in the string; assume UTC
-            return pytz.utc.localize(parsed)
+            return parse_into_datetime(value, self.precision)
+        except ValueError:
```
gtback: I don't think we need the `try`/`except`/`raise` here (we aren't rewriting or adding any additional information to the Exception). Unless I'm missing something.

stix2/test/conftest.py
```diff
@@ -13,12 +12,12 @@
 @pytest.fixture
 def clock(monkeypatch):
 
-    class mydatetime(dt.datetime):
+    class mydatetime(stix2.utils.STIXdatetime):
```
gtback: There might be easier ways to patch our custom datetime subclass. I remember some of these shenanigans were due to the fact that `dt.datetime` is built in. We can leave this as is, though, since it should work.

stix2/test/test_campaign.py
```diff
@@ -23,8 +23,8 @@ def test_campaign_example():
     campaign = stix2.Campaign(
         id="campaign--8e2e2d2b-17d4-4cbf-938f-98ee46b3cd3f",
         created_by_ref="identity--f431f809-377b-45e0-aa1c-6a4751cae5ff",
-        created="2016-04-06T20:03:00Z",
-        modified="2016-04-06T20:03:00Z",
+        created="2016-04-06T20:03:00.000Z",
```
gtback: Just to be clear, we shouldn't *need* to pass in the microsecond precision for the constructor args, correct? I'm fine doing it in most cases, but there should be at least one test to confirm it's not required.

stix2/utils.py
```diff
@@ -12,15 +12,34 @@
...
+            kwargs['month'] = dttm.month
+            kwargs['day'] = dttm.day
+            kwargs['hour'] = dttm.hour
+            kwargs['minute'] = dttm.minute
+            kwargs['second'] = dttm.second
+            kwargs['microsecond'] = dttm.microsecond
+            kwargs['tzinfo'] = dttm.tzinfo
+        self = dt.datetime.__new__(cls, *args, **kwargs)
+        self.precision = precision
+        return self
```
gtback: I wish we didn't need to do this attribute-by-attribute copying. I knew there was a `timetuple()` method on `datetime` instances, but I just realized that it doesn't do exactly what we need, particularly with `tzinfo`. 

Is there any reason we couldn't just use (something like) the following?
```
     precision = kwargs.pop('precision', None)
     if isinstance(args[0], dt.datetime):  # Allow passing in a datetime object
         dttm = args[0]
     else:
         dttm = dt.datetime.__new__(cls, *args, **kwargs)
     dttm.precision = precision
     return dttm
```

stix2/utils.py
```diff
@@ -29,32 +48,47 @@ def format_datetime(dttm):
     else:
         zoned = dttm.astimezone(pytz.utc)
     ts = zoned.strftime("%Y-%m-%dT%H:%M:%S")
-    if zoned.microsecond > 0:
-        ms = zoned.strftime("%f")
+    ms = zoned.strftime("%f")
+    precision = getattr(dttm, "precision", None)
+    if precision:
+        if precision == "millisecond":
+            ts = ts + '.' + ms[:3]
```
gtback: I don't think we need both nested `if`s. In fact, we can probably replace lines 52-54 with `if getattr(dttm, "precision", None) == "millisecond":`

stix2/utils.py
```diff
@@ -29,32 +48,47 @@ def format_datetime(dttm):
...
+    elif zoned.microsecond > 0:
         ts = ts + '.' + ms.rstrip("0")
     return ts + "Z"
 
 
-def parse_into_datetime(value):
+def parse_into_datetime(value, precision=None):
```
gtback: I'd feel a bit better if we had some more tests to handle the additional complexity of this function.

PR #27, opened 2017-07-05
Allow passing a list to Bundle constructor

### LINKED ISSUE #26: Passing a list to Bundle constructor creates list within a list
Currently if you pass a list to the Bundle constructor it creates a list within a list: 

```json
        {
          "objects":
             [
               [
                   { },
                   { }
               ]
             ]
        }
```

If the first argument is a list, the constructor should just assign it to `objects`.

clenk:
Fix #26

### DIFF
--- a/stix2/bundle.py
+++ b/stix2/bundle.py
@@ -17,6 +17,9 @@ class Bundle(_STIXBase):
     def __init__(self, *args, **kwargs):
         # Add any positional arguments to the 'objects' kwarg.
         if args:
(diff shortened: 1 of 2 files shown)


gtback:
I was a little unsure about the order that objects are put into the list (originally the `objects` kwarg came first), but your test cases confirmed for me that your order makes the most sense. 👍

clenk:
@gtback The test cases are what led me to change the order. :)

PR #30, opened 2017-07-12
Merge in Pattern Expression stuff

rpiazza:


stix2/pattern_expressions.py
```diff
@@ -0,0 +1,223 @@
...
+
+class PatternExpression(object):
+
+    @staticmethod
+    def escape_quotes_and_backslashes(s):
+        return s.replace(u'\\', u'\\\\').replace(u"'", u"\\'")
+
+
+class ComparisonExpression(PatternExpression):
+    def __init__(self, operator, lhs, rhs, negated=False):
```
gtback: I think maybe `lhs, operator, rhs` makes more sense here?

stix2/pattern_expressions.py
```diff
@@ -0,0 +1,223 @@
...
+
+class BooleanExpression(PatternExpression):
+    def __init__(self, operator, operands):
+        self.operator = operator
+        self.operands = []
+        for arg in operands:
+            if not hasattr(self, "root_type"):
+                self.root_type = arg.root_type
+            elif self.root_type and (self.root_type != arg.root_type) and operator == "AND":
+                raise ValueError("This expression cannot have a mixed root type")
```
gtback: I think this error message would be clearer if it said something like `All operands to an "AND" expression must have the same object type`.

stix2/pattern_expressions.py
```diff
@@ -0,0 +1,223 @@
+from .constants import Constant, IntegerConstant, ListConstant, make_constant
+from .object_path import ObjectPath
+
+
+class PatternExpression(object):
```
gtback: We should probably use the name `_Expression` here, or maybe just make the staticmethod a function?

stix2/pattern_expressions.py
```diff
@@ -0,0 +1,223 @@
...
+    def __init__(self, operands):
+        super(AndBooleanExpression, self).__init__("AND", operands)
+
+
+class OrBooleanExpression(BooleanExpression):
+    def __init__(self, operands):
+        super(OrBooleanExpression, self).__init__("OR", operands)
+
+
+class ObservableExpression(PatternExpression):
```
gtback: Should be called `ObservationExpression` to match the spec.

stix2/pattern_expressions.py
```diff
@@ -0,0 +1,223 @@
...
+        return "(%s)" % self.expression
+
+
+class ExpressionQualifier(PatternExpression):
+    pass
+
+
+class RepeatQualifier(ExpressionQualifier):
+    def __init__(self, times_to_repeat):
+        if isinstance(times_to_repeat, IntegerConstant):
```
gtback: Should not use IntegerConstant here (or in `WithinQualifier` or `StartStopQualifier`. These are not constants in the same way that the RHS of a comparison expression is. Just an `int` should be fine.

stix2/pattern_expressions.py
```diff
@@ -0,0 +1,223 @@
...
+        elif isinstance(stop_time, int):
+            self.stop_time = IntegerConstant(stop_time)
+        else:
+            raise ValueError("%s is not a valid argument for a Within Qualifier" % stop_time)
+
+    def __str__(self):
+        return "START %s STOP %s" % (self.start_time, self.stop_time)
+
+
+class QualifiedObservationExpression(PatternExpression):
```
gtback: Is this the most general class? Can you have qualifiers as part of a sub-expression?

gtback:
In addition to my comments above, I'd like to see how we can make this API be more "friendly" and feel less like a Java-esque [Kingdom of Nouns](https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html). Having the classes is OK for verbose, explicit construction of patterns, but having more Pythonic APIs. I can try to come up with some examples for what I mean by that.

stix2/constants.py
```diff
@@ -0,0 +1,135 @@
...
+# TODO: REConstant?
+# TODO: Timestamp
+
+
+class Constant(object):
+    def __str__(self):
+        return "%s" % self.value
+
+    @staticmethod
+    def escape_quotes_and_backslashes(s):
```
rpiazza: Sure - I was in a static method mood

stix2/constants.py
```diff
@@ -0,0 +1,135 @@
...
+    "SHA3224": ("^[a-fA-F0-9]{56}$", "SHA3-224"),
+    "SHA3256": ("^[a-fA-F0-9]{64}$", "SHA3-256"),
+    "SHA3384": ("^[a-fA-F0-9]{96}$", "SHA3-384"),
+    "SHA3512": ("^[a-fA-F0-9]{128}$", "SHA3-512"),
+    "SSDEEP": ("^[a-zA-Z0-9/+:.]{1,128}$", "ssdeep"),
+    "WHIRLPOOL": ("^[a-fA-F0-9]{128}$", "WHIRLPOOL"),
+}
+
+
+class HashConstant(StringConstant):
```
rpiazza: Yes - I was working on something to determine what the constant should be based on the object path, as we had discussed - because I thought it would be relatively easy - but there are some inconsistencies with the Property classes - so I decided to go with stuff like this.

stix2/object_path.py
```diff
@@ -0,0 +1,59 @@
...
+        self.property_path.extend(other.property_path)
+        return self
+
+    @staticmethod
+    def make_object_path(lhs):
+        path_as_parts = lhs.split(":")
+        return ObjectPath(path_as_parts[0], path_as_parts[1].split("."))
+
+    @staticmethod
+    def create_ObjectPathComponent(component_name):
```
rpiazza: sure

stix2/pattern_expressions.py
rpiazza: Actually, there are a lot of classes that shouldn't be visible - so I changed them to use the "_" notation

stix2/pattern_expressions.py
rpiazza: ok

rpiazza: actually, its a hidden class - so it doesn't matter

stix2/pattern_expressions.py
rpiazza: nice

stix2/constants.py
gtback: Ha, I get into those moods occasionally. Now I'm in a "make everything a module-level function" mood.

stix2/pattern_expressions.py
rpiazza: Done

stix2/pattern_expressions.py
rpiazza: Not sure I agree - These are constants of the Pattern Expression, not STIX data types.

stix2/pattern_expressions.py
rpiazza: If I think I understand your question - see test case: test_multiple_qualifiers

stix2/pattern_expressions.py
rpiazza: You don't create instances of ComparisonExpression, but EqualityComparisonExpression, MatchesComparisonExpression, etc.  This is to enable the PatternMatcher to have different functionality for each type of ComparisonExpression - assuming we start using this stuff in Andy's code...

stix2/constants.py
```diff
@@ -0,0 +1,135 @@
+import base64
```
rpiazza: I think rolling up to stix2.patterns is the right level...

I will make this change.

gtback:
Closing in favor of #35.

PR #34, opened 2017-07-17
Object factory

### LINKED ISSUE #28: Create an Object Factory class
This would let users specify defaults for some common properties and easily create STIX objects without needing to specify them every single time; the objects would automatically be created with the default value.

For example, you could specify a default `created_by_ref` value to point to the Identity SDO for your organization.

clenk:
This is an initial implementation of the object factory, which allows users to define defaults for some common properties which will be applied to all STIX objects. Other properties (granular markings, for example) can be added in the future.

Close #28.

### DIFF
--- a/stix2/properties.py
+++ b/stix2/properties.py
@@ -5,7 +5,7 @@ import inspect
 import re
 import uuid
 
-from six import text_type
+from six import string_types, text_type
 
 from .base import _Observable, _STIXBase
 from .exceptions import DictionaryKeyError
@@ -101,12 +101,9 @@ class ListProperty(Property):
             iter(value)
         except TypeError:
             raise ValueError("must be an iterable.")
-        try:
-            if isinstance(value, _STIXBase) or isinstance(value, basestring):
-                value = [value]
-        except NameError:
-            if isinstance(value, str):
-                value = [value]
+
+        if isinstance(value, (_STIXBase, string_types)):
+            value = [value]
 
         result = []
         for item in value:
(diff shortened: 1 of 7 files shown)


stix2/properties.py
```diff
@@ -102,7 +102,7 @@ def clean(self, value):
         except TypeError:
             raise ValueError("must be an iterable.")
         try:
-            if isinstance(value, basestring):
+            if isinstance(value, _STIXBase) or isinstance(value, basestring):
```
gtback: `isinstance()` can take a tuple as the second argument, so:

```python
if isinstance(value, (_STIXBase, basestring)):
```

chisholm:
The default lists need to be deep-copied before being modified, otherwise they can keep growing with every object you create (if you choose to append values).  E.g. [this](https://github.com/oasis-open/cti-python-stix2/blob/0a2cda00cc41a7d29ae7c20e0b923c0dd7298330/stix2/environment.py#L47) is a shallow copy.  So [this](https://github.com/oasis-open/cti-python-stix2/blob/0a2cda00cc41a7d29ae7c20e0b923c0dd7298330/stix2/environment.py#L58) will actually change the default, affecting future object creation.

chisholm:
If appending to list-valued defaults is to be supported, it seems reasonable to support appending multiple values at a time.  E.g. `factory.create(SdoClass, external_references=[ref1, ref2])`.  As written though, that will result in a nested list, e.g. `[def1, def2, [ref1, ref2]]`.  Should appending multiple values be supported?

gtback:
Good catch, @chisholm . I realized I was doing a shallow copy when I wrote that line, but figured it didn't matter since (at the time) values would only get overwritten, not modified. Then I asked @clenk  to make them modifiable, forgetting that the shallow copy was no longer sufficient. I should have put a comment about the shallow copy.

PR #35, opened 2017-07-18
Pattern expressions

gtback:
Add functionality to make it easier to construct pattern expressions.


gtback:
The comments in #30 still need to be addressed.

PR #36, opened 2017-07-18
Data stores

gtback:
Refactoring of Data Stores (and the component Sources and Sinks), and the addition of FileSystem and memory stores.

TODO:
* [x] Remove duplicate code between `CompositeDataSource` and `DataSource`, and make `CompositeDataSource` a subclass of `DataSource`
* [x] Remove duplicate code in `memory.py`
* [x] Ensure filters on `CompositeDataSource` get passed to child `DataSource`s correctly.
* [x] Review `STIXCommonPropertyFilters` for idiomatic Python
* [x] `deduplicate` can be made more idiomatic with a set()
* [x] Consider changing filters from dicts to `namedtuples`
* [x] Add tests and ensure 95%+ test coverage.


rpiazza:
A fair amount of the code in memory.py seems to be duplicated, or very similar...

chisholm:
I was surprised that CompositeDataSource isn't a kind of DataSource (i.e. a subclass).  The CompositeDataSource indicates there can be only one per "workspace" (if I'm understanding the thinking correctly).  Perhaps this is supposed to be the way workspaces manage datasources.  If there needs to be some layer between the workspace and its datasource(s), this CompositeDataSource is named poorly.  It _really_ suggests a subtype of DataSource.

DataSource and CompositeDataSource have a lot of methods in common:
- get
- all_versions
- query
- add_filter
- remove_filter
- get_filters (why is this a getter method in DataSource, but a `@property` in CompositeDataSource?)
- deduplicate

So it feels to me like they could have a relationship.  add_filter and remove_filter seem like they were copy-pasted anyway.

Why do workspaces need this wrapping layer anyway?  Why shouldn't it be possible for the user to choose a single (i.e. non-composite) data source with his workspace?

(Again, apologies if I'm out of touch with the latest design thinking :))

stix2/sources/__init__.py
```diff
@@ -54,192 +59,191 @@ def make_id():
...
+class DataStore(object):
+    """
+    An implementer will create a concrete subclass from
+    this abstract class for the specific data store.
     """
+    def __init__(self, name="DataStore"):
+        self.name = name
+        self.id = make_id()
+        self.source = DataSource()
+        self.sink = DataSink()
```
chisholm: DataSource and DataSink are documented to be abstract; I don't think they should be instantiated.  If DataStore is itself an abstract base class, perhaps just setting the fields to None would work?

Also, you've created read-only properties named "source" and "sink" (because they don't have setters).  So you're assigning to a read-only property.  That will produce an error.

stix2/sources/__init__.py
```diff
@@ -557,150 +687,198 @@ def add_filter(self, filters):
...
         unique = []
-        have = False
+        dont_have = False
         for i in stix_obj_list:
+            dont_have = False
             for j in unique:
-                if i['id'] == j['id'] and i['modified'] == j['modified']:
-                    have = True
+                for field in STIX_VERSION_FIELDS:
+                    if not i[field] == j[field]:
```
chisholm: '!=' is much nicer than 'not =='.  And does your `break` line below need an extra level of indent?

If `i` and `j` have a mismatched field, that means the objects are different.  It doesn't mean `i` isn't in the `unique` list.  It could just be in a different location, i.e. it could match a different `j`.  I think DataSource.deduplicate does it right: it breaks on a match instead of a non-match.

Looks like you switched to using `STIX_VERSION_FIELDS` here, but not in DataSource?  Seems like there are two implementations that have lost sync.  Perhaps you could reuse a single implementation.  Neither method uses `self`, so they don't even have to be methods.  Maybe make it a free function?

stix2/sources/__init__.py
gtback: Yes, the `source` and `sink` properties should not be there. Also, the Abstract class should not be instantiated. `None` is a fine default.

gtback: Fixed by f4335f43ad280903afa392c880098206d71fdd35

gtback:
@rpiazza is correct; there is a lot of duplicate code that needs reviewed and probably removed.

This also needs some additional tests, and update the current tests to pass.

stix2/sources/__init__.py
```diff
@@ -557,150 +687,198 @@ def add_filter(self, filters):
...
+            return stix_obj_field > filter_["value"]
+        elif filter_["op"] == "<":
+            return stix_obj_field < filter_["value"]
+        elif filter_["op"] == ">=":
+            return stix_obj_field >= filter_["value"]
+        elif filter_["op"] == "<=":
+            return stix_obj_field <= filter_["value"]
+        else:
+            return -1
+
```
chisholm: Could all of these methods be staticmethods?  Anyway, making a class full of static methods seems like a Java idiom; why do it in Python?  Could these all be free functions?

stix2/sources/__init__.py
```diff
@@ -557,150 +687,198 @@ def add_filter(self, filters):
...
+    @classmethod
+    def _boolean(filter_, stix_obj_field):
+        if filter_["op"] == "=":
+            return stix_obj_field == filter_["value"]
+        elif filter_["op"] == "!=":
+            return stix_obj_field != filter_["value"]
+        else:
+            return -1
+
```
chisholm: These methods (_id and _boolean) look identical, modulo a variable renaming.  Could they be combined into one?

gtback:
@chisholm: I think the current CompositeDataSource implementation is from before the other classes were updated. I'm not really sure if there's value in having abstract DataSource and DataSink class. A more Pythonic way is for any class that implements the correct functions to work as a DataSource (or DataSink). An abstract DataStore probably makes sense, since the implementations of `get`, `all_versions`, `query`, and `save` just delegate to the component classes, and the concrete DataStore implementations can be much simpler.

CompositeDataSource should definitely implement the DataSource API. Its entire purpose it to stand in for a collection of other DataSource's, allowing federated object retrieval and querying.

stix2/sources/__init__.py
```diff
@@ -557,150 +687,198 @@ def add_filter(self, filters):
...
+    @classmethod
+    def _string(cls, filter_, stix_obj_field):
+        return cls._all(filter_, stix_obj_field)
+
+    @classmethod
+    def _timestamp(cls, filter_, stix_obj_timestamp):
+        return cls._all(filter_, stix_obj_timestamp)
+
```
chisholm: Same here: _string and _timestamp are the same.  Could they be combined?  Actually, they just call _all, which takes the exact same fields.  You could replace all uses of either of these methods with _all.  Did you envision a future use for the extra layer of indirection?

stix2/sources/memory.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        self.name = name
+        self.id = make_id()
+        self.data = {}
+
+        if stix_data:
+            if type(stix_data) == dict:
+                # stix objects are in a bundle
+                # verify STIX json data
+                r = validate_string(json.dumps(stix_data))
```
chisholm: This stringification seems silly, as stix-validator is just going to do a `json.loads` to put it back into Python structures again.  Perhaps we need a stix-validator function that works off of parsed JSON (would `schema_validate` work?  Its docstring says it needs a string, but the code clearly doesn't treat it that way.  Looks like the docstring needs an update.)

chisholm:
Regarding abstract DataSource and DataSink classes, I think duck typing with our own custom concepts is fine, but it can be trickier to define and document (and to stick to the plan!  Little things like equality checks, stringification, etc might unintentionally add requirements).  One nice thing about having an abstract class is that it can act as a central place to document the requirements, and might make it easier for someone to create a class which satisfies the requirements, whatever they may be.  You'd just have to explicitly say that users don't _have_ to subclass the abstract class; any value which satisfies the requirements will do.  The abstract class is only a convenience.

stix2/sources/memory.py
clenk: That is correct. I will change the docstring in the validator.

stix2/sources/__init__.py
```diff
@@ -54,192 +59,183 @@ def make_id():
...
+    this abstract class for the specific data store.
     """
+    def __init__(self, name="DataStore"):
+        self.name = name
+        self.id = make_id()
+        self.source = None
+        self.sink = None
 
-    def __init__(self, name="CompositeDataSource"):
+    def get(self, stix_id):
```
clenk: `_composite_filters` is listed as an arg in the docstring, but is absent here.

rpiazza:
Since all of the api methods on the DataStore classes just delegate to the source or sink, and that is defined in the DataStore abstract class - why are they redefined in all of the other DataStore classes??

stix2/sources/__init__.py
```diff
@@ -314,175 +310,301 @@ def add_filter(self, filters):
...
+                # not a common filter
+                if 'id' in filter_ and self.filter_allowed[filter_['id']] is False:
+                    continue
+
+                # check filter "field" is in STIX object - if cant be applied
+                # due to STIX object, STIX object is discarded (i.e. did not
+                # make it through the filter)
+                if filter_['field'] not in stix_obj.keys():
+                    clean = False
+                    break
```
clenk: Instead of setting `clean` to `False`, could you just do `continue` if you determine the object shouldn't be added to result? This could simplify the code.

stix2/sources/__init__.py
clenk: Having them in a class allows this line to work: https://github.com/oasis-open/cti-python-stix2/blob/611045528f6537bcd31327e0ec1f15866437645d/stix2/sources/__init__.py#L388

gtback:
@rpiazza: You are correct, those functions shouldn't be re-defined in the subclasses of `DataSource`

stix2/sources/__init__.py
chisholm: Ah, I hadn't seen that.  But I think that will work with `@staticmethod`.  Anyway, if I just needed to look up a function by name, I think I'd have just used a dict.  The underscore functions aren't ever called via that `getattr` line (right?), so it seems like they could be free functions.  Recently, Greg has made comments that some designs don't seem very "Pythonic".  This class full of static methods design really smacks to me of Java, i.e. doesn't seem Pythonic :)  Though you/@gtback may disagree.

clenk: Yeah, [functions should be organized into modules instead of classes](https://github.com/amontalenti/elements-of-python-style#functions-vs-classes). I like the dict approach, too - would make it easier for a user to define a filter type for their custom properties if necessary.

stix2/sources/__init__.py
clenk: I was also wondering about this. It feels tedious to me to require a separate function for each property. Could we have a function for each data type and then a mapping of property names to datatype functions? Then adding a new property would only require adding an entry to the mapping rather than defining a whole new function. Or is there a better way to handle it?

gtback:
I pushed some changes that at least get the tests passing. I consolidated a bunch of the TODOs in the description of this issue. Once those are done I think we can merge.

gtback:
D'oh. I forgot the TAXII client needs to be on PyPI before the tests will pass on Travis. I'll take care of that in the morning.

stix2/sources/__init__.py
clenk: Ah, no; I didn't see the double for-loop.

stix2/sources/__init__.py
chisholm: Type-specific checks...  Feels familiar: the matcher does that, since it has to check property values of particular types from a JSON object.  Similar considerations could come into play with these filters.  E.g. one could decide that if types are incompatible, that's a failed match but not an error, just like the pattern spec.  Might people ever want to filter in ways analogous to how patterns work?  E.g. have IP addresses and subnet tests?  Regex matches on strings?  Something to think about.

Filters are supposed to operate on our SDO python types, right?  Should one be able to determine property types from an SDO instance?  E.g. look up in _properties?  But perhaps _properties should be considered an implementation detail.  Could a method be written into _STIXBase to determine a type given a property name?  Then all you'd need is a type->tester_function mapping.

Although "identifier" is defined as a distinct type (section 2.5), I'm not sure that it merits its own type for filtering.  Wouldn't plain old string tests do the trick?  Is `STIXCommonPropertyFilters._id` necessary?  Would it be so bad to allow all of the string ops on an ID (even if they would probably mostly never be used)?

gtback:
To reply to my earlier comment, the TAXII client isn't quite ready to be put on PyPI yet, so this will have to wait until that's done. There's still a bunch of refactoring that we can do here, though.

stix2/sources/filesystem.py
```diff
@@ -0,0 +1,201 @@
...
+            # the specified type. Compile correct dir paths
+            for dir_ in os.listdir(self.stix_dir):
+                if os.path.abspath(dir_) not in declude_paths:
+                    include_paths.append(os.path.abspath(dir_))
+
+        # grab stix object ID as well - if present in filters, as
+        # may forgo the loading of STIX content into memory
+        if "id" in [filter_["field"] for filter_ in file_filters]:
+            for filter_ in file_filters:
+                if filter_["field"] == "id" and filter_["field"] == '=':
```
emmanvg: `and filter_["field"] == '=':` "field" should be "op" correct?

emmanvg: Also, should it break immediately if the condition is met?

chisholm: For the first, looks like a bug to me.  For the second, absent some kind of policy stating the last "id" filter should always be used, I'd say do the efficient thing and break when you find the first.  Should "!=" be  handled on "id" filters too?

emmanvg:
Also: https://github.com/oasis-open/cti-python-stix2/blob/datastores/stix2/sdo.py#L198 has potential to be problematic. `properties={}` gets the following message:

"This inspection detects when a mutable value as list or dictionary is detected in a default value for an argument. Default argument values are evaluated only once at function definition time, which means that modifying the default value of the argument will affect all subsequent calls of the function."

chisholm:
Aha, coincidentally the taxii server had this exact problem, which did result in some nonsensical behavior. :)  Definitely a good thing to detect!

gtback:
Yes, the mutable defaults thing is something we should try to avoid. In this case, it's probably ok, since we never actually modify `properties` in the body of the decorator. Though, on the other hand, it may cause problems if you call it multiple times, due to the "closure-style" nature of the decorator implementation.

I don't think that's part of this PR, though.

emmanvg:
I fixed the `properties` problem on this PR, and also issue #39. Right now, It is only failing because of the taxii2 client import...

stix2/sdo.py
```diff
@@ -225,12 +321,31 @@ def custom_builder(cls):
...
+            _properties.update([
+                ('revoked', BooleanProperty()),
+                ('labels', ListProperty(StringProperty)),
+                ('external_references', ListProperty(ExternalReference)),
+                ('object_marking_refs', ListProperty(ReferenceProperty(type="marking-definition"))),
+                ('granular_markings', ListProperty(GranularMarking)),
+            ])
+
+            # Put all custom properties at the bottom, sorted alphabetically.
+            _properties.update(sorted([x for x in properties if x[0].startswith("x_")], key=lambda x: x[0]))
```
clenk: Since the user is creating a custom SDO instead of adding custom properties to an official SDO, I don't think the non-common properties will start with "x_". (Even custom properties added to official SDOs may not, as it is a SHOULD requirement not a MUST.) Should it check a list of common property names instead?

clenk:
I'm curious about the removal of the COMMON_PROPERTIES dictionary. Did it not work with OrderedDicts?

emmanvg:
Hi Chris, unless we are using COMMON_PROPERTIES for another purpose (e.g. validation of some sort). That dictionary is no longer needed as every single SDO/SRO has those common properties built-in.

stix2/sdo.py
emmanvg: Sure, I could make those changes.

emmanvg: I guess that would re-introduce the COMMON_PROPERTIES dictionary and just add properties in the same order they were provided.

clenk:
My concern is that we're duplicating code. I guess it was removed because we want `type` and `id`, which must be defined on each individual type, to be at the top. Could we do something like this though (for observed data, for example):

```python
_properties = OrderedDict([
    ('type', TypeProperty(_type)),
    ('id', IDProperty(_type)),
])
_properties.update(COMMON_PROPERTIES)
_properties.update({
    'first_observed': TimestampProperty(required=True),
    'last_observed': TimestampProperty(required=True),
    'number_observed': IntegerProperty(required=True),
    'objects': ObservableProperty(),
})
```

I also noticed that `revoked`, `labels` (if present), ` external_references`, `object_marking_refs`, and `granular_markings` were put at the bottom. I assumed they would go with the other common properties, but I could be wrong. The example indicator in [Part 1, Section 5.2](https://docs.google.com/document/d/1dIrh1Lp3KAjEMm8o2VzAmuV0Peu-jt9aAh1IHrjAroM/pub#h.rvtdrdkf1jdv) has `object_marking_refs` before `name`, `description`, and `pattern`.

stix2/sdo.py
clenk: > custom properties starting with "x_" had to be at the bottom of the object...

Right. But here an entire custom type is being created, so I don't think any properties will start with "x_". My understanding is that "x_" is only if you are adding custom properties to existing, official SDOs.

gtback: Right. I suppose you could have custom properties on a custom object (if you're extending someone else's custom object), but those should be treated like the custom properties on any other object, and completely separate from the "defined" properties of the custom object itself. I think?

stix2/sources/__init__.py
```diff
@@ -146,14 +146,14 @@ class DataSink(object):
     different sink components.
 
     Attributes:
-        id (str): A unique UUIDv4 to identify this DataSink.
+        id_ (str): A unique UUIDv4 to identify this DataSink.
```
gtback: Is something (pycodestyle, flake8, etc.) complaining about our use these names (`id` and `filter`)? Even though we are shadowing builtins, they aren't builtins we're actually using, so IMO we shouldn't use trailing underscores (I regret having done this in python-stix). If we're using *keywords* (like `class`, `import`, `from`, `for`), then we'll need to use the trailing underscore to prevent syntax errors.

stix2/sources/filesystem.py
```diff
@@ -19,7 +19,7 @@
 class FileSystemStore(DataStore):
     """
     """
-    def __init__(self, stix_dir="stix_data", name="FileSystemStore"):
+    def __init__(self, name="FileSystemStore", stix_dir="stix_data"):
```
gtback: Why does this need a `name`? If we want to uniquely identify the Source/Sink/Store in a human-readable way, I'd say the absolute path to the data directory should be sufficient.

stix2/sources/filesystem.py
```diff
@@ -71,8 +71,8 @@ def stix_dir(self):
         return self.stix_dir
 
     @stix_dir.setter
-    def stix_dir(self, dir):
-        self.stix_dir = dir
+    def stix_dir(self, dir_):
+        self.stix_dir = dir_
```
gtback: Same argument here. I'd prefer `directory` to `dir_`. I'd also rather call this `data_dir` than `stix_dir`.

stix2/test/test_custom.py
```diff
@@ -128,6 +128,40 @@ def test_custom_observable_object():
         NewObservable(property1='something', property2=4)
 
 
+def test_custom_no_properties_raises_exception():
+    with pytest.raises(ValueError):
```
gtback: Will all five of these custom objects create errors? If so, there should be five separate tests. If not, the ones that don't should be outside the `pytest.raises` block (and also likely in their own tests anyway)

stix2/test/test_observed_data.py
```diff
@@ -173,7 +173,7 @@ def test_parse_observed_data(data):
     }""",
 ])
 def test_parse_artifact_valid(data):
-    odata_str = re.compile('"objects".+\}', re.DOTALL).sub('"objects": { %s }\n}' % data, EXPECTED)
+    odata_str = re.compile('\"objects\": {(?:.*?)(?:(?:[^{]*?)|(?:{[^{]*?}))*}', re.DOTALL).sub('"objects": { %s }' % data, EXPECTED)
```
gtback: If we're re-using the same regex, it should be defined outside these tests and re-used.

stix2/sources/memory.py
clenk: instead of
```python
r = validate_string(json.dumps(stix_data))
```
use the following:
```python
r = validate_instance(stix_data)
```
(and similarly for the other uses of `validate_string()`)

stix2/sources/memory.py
```diff
@@ -0,0 +1,297 @@
...
+
+        """
+        super(MemorySink, self).__init__(name=name)
+
+        if _store:
+            self.data = stix_data
+        else:
+            self.data = {}
+            if stix_data:
+                if type(stix_data) == dict:
```
clenk: This won't catch the case where stix_data is an instance of one of our SDO classes. From the docstring it looks like this might be intentional, but it also won't catch the case where stix_data is an instance of our Bundle class.

stix2/sources/filesystem.py
emmanvg: Why do we need a `name` for any of them?

stix2/sources/__init__.py
emmanvg: Alright. I thought it as an oversight, I'll revert that.

stix2/test/test_custom.py
emmanvg: Yep. I removed some of them in a later commit since they were covering the same area. I divided them into individual test cases and made sure the code coverage was good for each of the cases.

gtback:
I think this is good to go. The filtering code has been re-organized, and I'm not too worried about the drop in coverage (it's mostly due to the untested FileSystem source/sink). And all the tests are passing now that the TAXII client is on PyPI.

I'm going to merge this in, and we can handle further changes in new issues/branches.

PR #41, opened 2017-08-09
Use Namedtuples for Filters

gtback:
@emmanvg , @mbastian1135 , let me know if this makes sense. If so, we can merge into the datastores branch.


emmanvg:
I really like how it looks. Its more structured and more readable as well. Happy to merge this in. Any thoughts @mbastian1135 ?

PR #42, opened 2017-08-11
Parse bundles correctly

### LINKED ISSUE #40: Parsing Bundles is incomplete
`bundle` is not defined in OBJ_MAP, so calling `stix2.parse()` on a dict/JSON string representing a Bundle fails.

Also, the `objects` property of a Bundle does not convert dictionaries into actual python-stix2 objects. It should likely use a `ListProperty(STIXObject)` (where STIXObject is a class we haven't yet defined), or maybe a custom Property subclass, and re-use some of the logic from `parse` to detect the correct object type to construct.

clenk:
This required refactoring parts of the library. Code in \_\_init\_\_.py merged into bundle.py, which was renamed core.py. Code in other.py was merged into common.py.

Fixes #40.


gtback:
👍

Issue #29, opened 2017-07-12
No error when adding Granular Marking with invalid selector


clenk:
The API should throw an error if a granular marking is added to an object but one of its selectors references a property or list item that is not present on that object. Or if it doesn't throw an error, it should at least omit the invalid selector.

emmanvg:
Hi @clenk, @gtback ! I started an effort to work on this and on a simpler way to apply markings to objects. Though, I never finished and probably won't be able to in a couple of weeks. So, if someone is able to take that step. It would be awesome. I pushed my latest changes on my [fork](https://github.com/emmanvg/cti-python-stix2/tree/marking-support) (which might be outdated). Is currently [WIP], but hopefully will give some ideas on how to raise the error if you or someone else want to tackle this issue.

clenk:
Thanks, @emmanvg! I'll take a look.

PR #46, opened 2017-08-24
Marking Support

emmanvg:
Fixes issue #29. Successful integration of marking-prototype source code.


stix2/markings/__init__.py
```diff
@@ -0,0 +1,214 @@
...
+Python STIX 2.0 Data Markings API.
+
+These high level functions will operate on both object level markings and
+granular markings unless otherwise noted in each of the functions.
+"""
+
+from stix2.markings import granular_markings, object_markings
+
+
+def get_markings(obj, selectors, inherited=False, descendants=False):
```
gtback: `selectors` should probably be a kwarg defaulting to `None`

stix2/markings/__init__.py
```diff
@@ -0,0 +1,214 @@
...
+        descendants
+    )
+
+    if inherited:
+        results.extend(object_markings.get_markings(obj))
+
+    return list(set(results))
+
+
+def set_markings(obj, selectors, marking):
```
gtback: `selectors` should be kwarg defaulting to `None`. That means the order of args will have to change, which I'm not sure that I like, as it makes the function signatures somewhat inconsistent.

stix2/markings/granular_markings.py
```diff
@@ -0,0 +1,269 @@
...
+from stix2 import exceptions
+from stix2.markings import utils
+
+
+def get_markings(obj, selectors, inherited=False, descendants=False):
+    """
+    Get all markings associated to with the properties.
+
+    Args:
+        obj: An SDO or SRO object.
```
gtback: This should also work on dictionaries. I think this function already does.

stix2/markings/granular_markings.py
```diff
@@ -0,0 +1,269 @@
...
+    if not any(marking in granular_markings for marking in remove):
+        raise exceptions.MarkingNotFoundError(obj, remove)
+
+    granular_markings = [
+        m for m in granular_markings if m not in remove
+    ]
+
+    granular_markings = utils.compress_markings(granular_markings)
+
+    if granular_markings:
```
gtback: In order to make this work on dictionaries, we might need to reimplement `new_version` (or not, since in general our attitude is to trust people who build their own dictionaries to do the right thing).

stix2/test/constants.py
```diff
@@ -36,6 +45,7 @@
 MALWARE_KWARGS = dict(
     labels=['ransomware'],
     name="Cryptolocker",
+    description="A ransomware related to ..."
```
gtback: Is `description` required? I'd rather keep these a minimal set, but it would be fine to have an additional set for a "realistic" Malware object.

gtback:
I was having test failures locally, but it turned out I didn't have the latest copy of the code, so now that I have it, all the tests are passing for me, too.

stix2/markings/__init__.py
clenk: Is it more important to have consistent signatures or to be pythonic? We could make `marking` a kwarg with a default of `None` as well, and then check and throw an error if `marking` is `None`.

gtback: It's actually more consistent to have `markings` come before `selectors` (if you compare across all three files, especially `object_markings.py`) on functions that need to accept markings (not `get`, `clear`, or `is_marked`). I was too focused on making selectors the first argument, rather than ensuring consistent arguments for the same function across all modules. 

Here's what I'm thinking:

`markings/__init__.py`
```python
def get_markings(obj, selectors=None, inherited=False, descendants=False):
def set_markings(obj, marking, selectors=None):
def remove_markings(obj, marking, selectors=None):
def add_markings(obj, marking, selectors=none):
def clear_markings(obj, selectors=None):
def is_marked(obj, marking=None, selectors=None, inherited=False, descendants=False):
```

`markings/object_markings.py`
```python
def get_markings(obj):
def add_markings(obj, marking):
def remove_markings(obj, marking):
def set_markings(obj, marking):
def clear_markings(obj):
def is_marked(obj, marking=None):
```

`markings/granular_markings.py`
```python
def get_markings(obj, selectors, inherited=False, descendants=False):
def set_markings(obj, marking, selectors):
def remove_markings(obj, marking, selectors):
def add_markings(obj, marking, selectors):
def clear_markings(obj, selectors):
def is_marked(obj, marking=None, selectors, inherited=False, descendants=False):
```

The last one is tricky, because it obviously isn't valid. `selectors` is 'required' for `granular_markings.is_marked`, but would it make sense to give it a default value (in all of the `granular_markings` functions, of `[""]`?

clenk: I like this. For `granular_markings.is_marked()`, I think we can assume that if users don't provide a value for `selectors` then it is not what they intended to do and should raise an error. If they don't provide a selector to `markings.is_marked` it will call the `object_markings` version, so if this function is called and no selector is provided, they are calling the `granular_markings` version directly and probably just forgot to add that arg. The same is true for the other `granular_markings` functions, but by defining them as required args, Python will raise those errors. For `is_marked`, I think we should keep the argument order consistent, add a check, and raise the error ourselves.

I don't think `[""]` has any advantage over `None` as a default, and `None` will be easier for code maintainers to understand its purpose.

stix2/markings/granular_markings.py
gtback: This has been fixed.

stix2/test/constants.py
gtback: This has been fixed

gtback:
@emmanvg, @clenk: anything else that needs to be done before merging this?

clenk:
@gtback No, I think it's good to go.

emmanvg:
@gtback I think it's good too!

PR #47, opened 2017-08-24
Add support for custom extensions on Cyber Observables

clenk:
Also increase code coverage.

### DIFF
--- a/stix2/observables.py
+++ b/stix2/observables.py
@@ -768,12 +768,12 @@ def CustomObservable(type='x-custom-observable', properties={}):
             }
             # Check properties ending in "_ref/s" are ObjectReferenceProperties
             for prop_name, prop in properties.items():
-                if prop_name.endswith('_ref') and prop.__class__.__name__ != 'ObjectReferenceProperty':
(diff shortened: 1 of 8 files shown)


gtback:
@clenk: feel free to merge this, if my question is wrong or doesn't make sense.

stix2/observables.py
```diff
@@ -750,6 +766,16 @@ class _Custom(cls, _Observable):
...
                 'type': TypeProperty(_type),
             }
+            # Check properties ending in "_ref/s" are ObjectReferenceProperties
+            for prop_name, prop in properties.items():
+                if prop_name.endswith('_ref') and prop.__class__.__name__ != 'ObjectReferenceProperty':
+                    raise ValueError("'%s' is named like an object reference property but "
+                                     "is not an ObjectReferenceProperty." % prop_name)
+                elif (prop_name.endswith('_refs') and (prop.__class__.__name__ != 'ListProperty'
+                                                       or prop.contained.__class__.__name__ != 'ObjectReferenceProperty')):
+                    raise ValueError("'%s' is named like an object reference property but "
```
gtback: Do we want this to be different from the error message above ("is named like an object reference *list* property")?

PR #52, opened 2017-09-08
Environment layer

### LINKED ISSUE #37: Create "Environment" class
This is an abstraction that will encompass a data source, data sink, and object factory, and potentially other state, in a unified interface for interacting with external STIX repositories.

clenk:
Open questions:

- Should each Environment object keep track of its own mapping of custom object types and classes? The current implementation does not.
- The stix2 validator is currently not run on data in any of the datastore implementations. Should this be brought back? One thought is that the API should make it difficult to create invalid STIX anyway, so this would be redundant. On the other hand, the API may not catch all errors, and trying to do so may result in duplicate code.

Closes #37

### DIFF
--- a/stix2/environment.py
+++ b/stix2/environment.py
@@ -1,11 +1,10 @@
 import copy
 
+from .sources import CompositeDataSource
+
 
 class ObjectFactory(object):
-    """Object Factory
-
-    Used to easily create STIX objects with default values for certain
-    properties.
+    """Easily create STIX objects with default values for certain properties.
 
     Args:
         created_by_ref: Default created_by_ref value to apply to all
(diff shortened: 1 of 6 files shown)


stix2/environment.py
```diff
@@ -66,3 +73,84 @@ def create(self, cls, **kwargs):
...
+    def create(self, *args, **kwargs):
+        return self.factory.create(*args, **kwargs)
+    create.__doc__ = ObjectFactory.create.__doc__
+
+    def get(self, *args, **kwargs):
+        try:
+            return self.source.get(*args, **kwargs)
+        except AttributeError:
+            raise AttributeError('Environment has no data source to query')
+    get.__doc__ = DataStore.get.__doc__
```
gtback: I feel like these should use the `DataSource` docstrings, but after taking a look at those, they could probably be cleaned up and reorganized between `DataStore` and `DataSource`. It's fine for now. I created #53 to address that.

gtback:
My first impression is that keeping track of custom types should be an Environment function. A user could conceivably want different settings in different parts of your application, which is the purpose of the Environment class. If we decide that we want to enable custom types outside of an explicit environment (which I'm not sure we do), that means we need to keep an implicit environment for that. The core `parse` function would then call the parse function for that environment. An implicit environment is also a step towards the "Workbench" API, so maybe we just defer the "implicit custom types" to that level, and require an explicit Environment otherwise.

I'm fine with not running the validator for every DataSource/DataSink interaction. If it's needed, we can later add a boolean `validate` flag when constructing them, but it shouldn't be the default. In production applications where you trust how the STIX content is created, it's just unnecessary performance overhead, and you can always implement validation yourself next to any calls to DataSource/DataSink operations.

I'd love to hear others' opinions, especially if I'm missing any use cases.

clenk:
I agree that users will want different _settings_ in different parts of their application. But specifically for custom types, my thinking was that the only time you would have a problem with registering them "globally" (like we do currently) is if you need 2 with the same name but different properties. In that case you would need to be able to register them to separate environments, but I assumed that is an unlikely occurrence.

gtback:
I was also thinking that you might want to allow custom types in some environments but not others. That does bring up the questions of how to register the same custom types in multiple environments, whether to still have a set of "global" custom types that apply to all environments, whether environments should allow nesting/inheritence, etc. It gets pretty complicated and IMO over-engineered pretty quick.

I guess what I'm saying, though, is that *which* custom types are supported *is* an environment-specific setting, but I could be wrong. The fact that the `Environment.parse` function doesn't depend on anything else in the `Environment` is, to me, a sign that it doesn't belong there, or that there should be things it depends on.

PR #57, opened 2017-09-22
Update to Data* suite, FileSystem, Memory, TAXII and Filters

mbastian1135:
1. Updated docstrings in Data* suite, CompositeDataSource, FileSystem* suite, Memory* suite, TAXIICollection* suite
2. Modified the format with which STIX data can be pushed and retrieved from Data* suites. 
    * FileSystemStore/Sink, MemoryStore/Sink, TAXIICollectionStore/Sink can now be supplied with STIX data in any of the following forms: python STIX objects, dicts,  json-encoded strings, or a list of any of the previous forms
    * FileSystemStore/Source will default to returning python STIX objects (i.e. it will json load STIX object files, and then use parse() to get a python STIX object from that)
    * MemoryStore/Source will return STIX objects in the manner they were added to Memory - but MemoryStore/Sink restricts forms to either python STIX objects or dicts (converting other forms to one of those before adding)
    * TAXIICollectionStore/Source will return python STIX objects, converting TAXII returned dicts to python STIX objects (I know this was still being debated, and be changed, just did this way for now)
3. Tightened the way filters are passed around throughout the API layers. Essentially any filters attached to a DataSource are added to a set and maintained in a set throughout operation. While run time queries (which are also groups of filters) are provided as lists and passed as lists until they reach DataSource.query() where the query list will be converted to a set and then combined with the filter sets of that DataSource and any possible parent CompositeDataSource(s).
4. Updated test_data_sources so would not break. However, need to add tests for 2.

### DIFF
--- a/stix2/sources/__init__.py
+++ b/stix2/sources/__init__.py
@@ -7,12 +7,8 @@ Classes:
     DataSource
     CompositeDataSource
 
-TODO:Test everything
-
 Notes:
-    add_filter(), remove_filter(), deduplicate() - if these functions remain
-    the exact same for DataSource, DataSink, CompositeDataSource etc... -> just
-    make those functions an interface to inherit?
+    Q: We have add_filters() but no remove_filter()
 
 """
 
@@ -28,14 +24,100 @@ def make_id():
     return str(uuid.uuid4())
 
 
-class DataStore(object):
+def apply_common_filters(stix_objs, query):
+    """Evaluate filters against a set of STIX 2.0 objects.
+
+    Supports only STIX 2.0 common property fields
+
+    Args:
+        stix_objs (list): list of STIX objects to apply the query to
+
+        query (set): set of filters (combined form complete query)
+
+    Returns:
+        (list): list of STIX objects that successfully evaluate against
+            the query.
+
     """
+    filtered_stix_objs = []
+
+    # evaluate objects against filter
+    for stix_obj in stix_objs:
+        clean = True
+        for filter_ in query:
+            # skip filter as filter was identified (when added) as
+            # not a common filter
+            if filter_.field not in STIX_COMMON_FIELDS:
+                raise ValueError("Error, field: {0} is not supported for filtering on.".format(filter_.field))
+
+            # For properties like granular_markings and external_references
+            # need to break the first property from the string.
+            if "." in filter_.field:
+                field = filter_.field.split(".")[0]
+            else:
+                field = filter_.field
+
+            # check filter "field" is in STIX object - if cant be
+            # applied due to STIX object, STIX object is discarded
+            # (i.e. did not make it through the filter)
+            if field not in stix_obj.keys():
+                clean = False
+                break
+
+            match = STIX_COMMON_FILTERS_MAP[filter_.field.split('.')[0]](filter_, stix_obj)
+            if not match:
+                clean = False
+                break
+            elif match == -1:
+                raise ValueError("Error, filter operator: {0} not supported for specified field: {1}".format(filter_.op, filter_.field))
+
+        # if object unmarked after all filters, add it
+        if clean:
+            filtered_stix_objs.append(stix_obj)
+
+    return filtered_stix_objs
+
+
+def deduplicate(stix_obj_list):
+    """Deduplicate a list of STIX objects to a unique set
+
+    Reduces a set of STIX objects to unique set by looking
+    at 'id' and 'modified' fields - as a unique object version
+    is determined by the combination of those fields
+
+    Args:
+        stix_obj_list (list): list of STIX objects (dicts)
+
+    Returns:
+        A list with a unique set of the passed list of STIX objects.
+
+    """
+    unique_objs = {}
+
+    for obj in stix_obj_list:
+        unique_objs[(obj['id'], obj['modified'])] = obj
+
+    return list(unique_objs.values())
+
+
+class DataStore(object):
+    """DataStore
+
     An implementer will create a concrete subclass from
-    this abstract class for the specific data store.
+    this class for the specific DataStore.
+
+    Args:
+        source (DataSource): An existing DataSource to use
+             as this DataStore's DataSource component
+
+        sink (DataSink): An existing DataSink to use
+             as this DataStore's DataSink component
 
     Attributes:
         id (str): A unique UUIDv4 to identify this DataStore.
-        source (DataStore): An object that implements DataStore class.
+
+        source (DataSource): An object that implements DataSource class.
+
         sink (DataSink): An object that implements DataSink class.
 
     """
@@ -47,14 +129,13 @@ class DataStore(object):
     def get(self, stix_id):
         """Retrieve the most recent version of a single STIX object by ID.
 
-        Notes:
-            Translate API get() call to the appropriate DataSource call.
+        Translate get() call to the appropriate DataSource call.
 
         Args:
-            stix_id (str): the id of the STIX 2.0 object to retrieve.
+            stix_id (str): the id of the STIX object to retrieve.
 
         Returns:
-            stix_obj (dictionary): the single most recent version of the STIX
+            stix_obj: the single most recent version of the STIX
                 object specified by the "id".
 
         """
@@ -63,15 +144,13 @@ class DataStore(object):
     def all_versions(self, stix_id):
         """Retrieve all versions of a single STIX object by ID.
 
-        Implement:
-            Translate all_versions() call to the appropriate DataSource call
+        Implement: Translate all_versions() call to the appropriate DataSource call
 
         Args:
-            stix_id (str): the id of the STIX 2.0 object to retrieve.
+            stix_id (str): the id of the STIX object to retrieve.
 
         Returns:
-            stix_objs (list): a list of STIX objects (where each object is a
-                STIX object)
+            stix_objs (list): a list of STIX objects
 
         """
         return self.source.all_versions(stix_id)
@@ -79,17 +158,15 @@ class DataStore(object):
     def query(self, query):
         """Retrieve STIX objects matching a set of filters.
 
-        Notes:
-            Implement the specific data source API calls, processing,
-            functionality required for retrieving query from the data source.
+        Implement: Specific data source API calls, processing,
+        functionality required for retrieving query from the data source.
 
         Args:
             query (list): a list of filters (which collectively are the query)
                 to conduct search on.
 
         Returns:
-            stix_objs (list): a list of STIX objects (where each object is a
-                STIX object)
+            stix_objs (list): a list of STIX objects
 
         """
         return self.source.query(query=query)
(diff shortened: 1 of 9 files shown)


stix2/sources/__init__.py
```diff
@@ -136,194 +221,122 @@ def add(self, stix_objs):
...
 
         """
-        # check filter field is a supported STIX 2.0 common field
-        if filter.field not in STIX_COMMON_FIELDS:
-            raise ValueError("Filter 'field' is not a STIX 2.0 common property. Currently only STIX object common properties supported")
-
-        # check filter operator is supported
-        if filter.op not in FILTER_OPS:
-            raise ValueError("Filter operation (from 'op' field) not supported")
+        if isinstance(filters, list) or isinstance(filters, set):
```
gtback: Why get rid of the `add_filter`/`add_filters` distinction? It seems a bit circuitous to have a check every time through the function of whether it's a list or not. This approach does allow you to pass in a list-of-lists-of-filters, but I'm not sure we want/need that complexity.

stix2/sources/filesystem.py
```diff
@@ -27,87 +42,181 @@ def __init__(self, stix_dir="stix_data"):
...
+    to file directory of STIX objects.
+
+    Can be paired with a FileSystemSource, together as the two
+    components of a FileSystemStore.
+
+    Args:
+        stix_dir (str): path to directory of STIX objects
+
     """
     def __init__(self, stix_dir="stix_data"):
```
gtback: Does it make sense to have a default value for this kwarg? I know it isn't part of this change, but I just noticed it.

stix2/test/test_data_sources.py
```diff
@@ -212,39 +213,39 @@ def test_add_get_remove_filter(ds):
         Filter('created', '=', object()),
     ]
 
-    assert len(ds.filters) == 0
+    assert len(ds._filters) == 0
```
gtback: The fact that this information is needed outside the class means we should probably not make it private.

stix2/sources/__init__.py
```diff
@@ -28,14 +24,100 @@ def make_id():
     return str(uuid.uuid4())
 
 
-class DataStore(object):
+def apply_common_filters(stix_objs, query):
```
gtback: Does this function now make more sense in the filters.py file?

stix2/sources/__init__.py
```diff
@@ -28,14 +24,100 @@ def make_id():
...
+
+        # if object unmarked after all filters, add it
+        if clean:
+            filtered_stix_objs.append(stix_obj)
+
+    return filtered_stix_objs
+
+
+def deduplicate(stix_obj_list):
+    """Deduplicate a list of STIX objects to a unique set
```
gtback: Does this function make sense outside of sources? Maybe just put in the `stix2.utils` module?

gtback:
Overall, I really like the direction this is going. There are a few suggestions I have above. I'm also interested in getting @clenk's feedback. (The tests need to pass, too 😀 )

stix2/sources/__init__.py
mbastian1135: Yes

stix2/sources/__init__.py
```diff
@@ -136,194 +221,122 @@ def add(self, stix_objs):
...
+        functionality required for retrieving data from the data source
 
         Args:
             stix_id (str): the id of the STIX 2.0 object to retrieve. Should
                 return a single object, the most recent version of the object
                 specified by the "id".
 
-            _composite_filters (list): list of filters passed along from
-                the Composite Data Filter.
+            _composite_filters (set): set of filters passed from the parent
```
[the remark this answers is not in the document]
mbastian1135: Well within the DataSource.get() call, only a STIX ID should be supplied. Thus any filters applied to the call are those attached to the DataSource or supplied from a CompositeDataSource.

If you pass in filters to a call, then that is a query (i.e. use query())

(or am I missing something?)

stix2/sources/__init__.py
mbastian1135: That was something I was going to bring up. I didn't like having the distinction, just based on other APIs I have used they have had one method to do functionality like this.  Im okay with it either way, not that critical.

To your latter point, yep noticed that. Obviously I dont think anyone would use the method that way. I just liked that you could supply a list, set or Filter to one method and it just handles it all. Alternatively, you could just make the method accept a list, or go back to two distinct methods.

stix2/sources/filesystem.py
mbastian1135: Nah, that was just there from the beginning.

stix2/test/test_data_sources.py
mbastian1135: Even if just for test purposes?

mbastian1135:
Chris, feel free to enlighten me on how I can check what tests it is breaking. I only checked it against (and modified) test_data_sources.py

stix2/sources/__init__.py
```diff
@@ -28,14 +24,100 @@ def make_id():
...
+            match = STIX_COMMON_FILTERS_MAP[filter_.field.split('.')[0]](filter_, stix_obj)
+            if not match:
+                clean = False
+                break
+            elif match == -1:
+                raise ValueError("Error, filter operator: {0} not supported for specified field: {1}".format(filter_.op, filter_.field))
+
+        # if object unmarked after all filters, add it
+        if clean:
+            filtered_stix_objs.append(stix_obj)
```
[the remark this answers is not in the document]
mbastian1135: okay

stix2/sources/__init__.py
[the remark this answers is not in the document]
clenk: Would it make sense to use `self.filters` then, instead of passing them as an argument to the function?

stix2/sources/__init__.py
clenk: I like the idea of a single function that handles all cases, and that is what's been done in the `markings` modules. However, [this StackOverflow answer](https://softwareengineering.stackexchange.com/a/296765/) suggests consensus is to use two functions, the plural version calling the singular.

I don't care too much about this either, but regardless we should be consistent throughout the library as to whether we have both a singular and a plural function, or just a single function for both cases.

clenk:
If you look on Travis, there's 2 environment tests that are failing because there is now no `add_filter()`. So they would pass if you switched them to `add_filters()`.

stix2/sources/taxii.py
```diff
@@ -10,66 +10,125 @@
...
+            if stix_data["type"] == "bundle":
+                bundle = dict(stix_data)
+            else:
+                bundle = dict(Bundle(stix_data))
+
+        self.collection.add_objects(bundle)
 
     @staticmethod
     def create_bundle(objects):
+        """TODO: Remove?"""
```
clenk: Yes, this should be removed - it duplicates the `Bundle` constructor.

stix2/test/test_data_sources.py
clenk: It may be useful for users to be able to check the current list of filters.

stix2/sources/__init__.py
```diff
@@ -7,12 +7,8 @@
     DataSource
     CompositeDataSource
 
-TODO:Test everything
-
 Notes:
-    add_filter(), remove_filter(), deduplicate() - if these functions remain
-    the exact same for DataSource, DataSink, CompositeDataSource etc... -> just
-    make those functions an interface to inherit?
+    Q: We have add_filters() but no remove_filter()
```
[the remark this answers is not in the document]
gtback: This is why I had originally exposed `self.filters` as public. Users could modify/remove if they wanted. 

Now I'm thinking about adding a FilterSet class, which would do validation on adding, still support removal and modification, and we can get rid of `add_filter`/`add_filters` entirely.

stix2/sources/__init__.py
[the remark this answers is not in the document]
gtback: I think the issue is for when the filters are defined on a "parent" CompositeDataSource, but also need to be applied to queries to child data sources.

The FileSystem and Memory data stores both just delegate calls to `get` to `query`. Maybe the right approach is to do something similar for TAXII (and make the case where the only filter is an `id=` a special case that calls `get_object` rather than `get_objects`), and then `get` only needs to be implemented on the base class as a proxy to `query`.

stix2/sources/__init__.py
```diff
@@ -340,62 +353,67 @@ def __init__(self):
         self.data_sources = {}
```
clenk: `name` was removed so it should be taken out of the docstring, too.

stix2/sources/__init__.py
```diff
@@ -28,14 +24,100 @@ def make_id():
...
+
+    return filtered_stix_objs
+
+
+def deduplicate(stix_obj_list):
+    """Deduplicate a list of STIX objects to a unique set
+
+    Reduces a set of STIX objects to unique set by looking
+    at 'id' and 'modified' fields - as a unique object version
+    is determined by the combination of those fields
```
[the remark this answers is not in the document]
mbastian1135: For your latter comment, how would that be applied in CompositeDataSource? Wouldnt you need CompositeDataSource.DataSources to be an ordered dict as well then?

stix2/sources/__init__.py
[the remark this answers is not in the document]
mbastian1135: If you do the FilterSet class, would all Filter creation have to be done through that interface? Run time queries too, which are technically "set(s) of filters"? 

This mayalso solve another possible issue in apply_common_filters() where the supplied set of filters is always format checked before being applied - even if the filters come from a DataStore/Source or CompositeDataSource, where they are already checked when attached to them.

stix2/sources/__init__.py
mbastian1135: Will relocate

stix2/sources/__init__.py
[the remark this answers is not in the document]
mbastian1135: Thats how TAXIICollectionSource.get() currently is. It just adds in any filters(attached to the Source or from CompositeDataSource)

stix2/sources/taxii.py
mbastian1135: removed

stix2/sources/__init__.py
mbastian1135: good catch, removed

stix2/sources/__init__.py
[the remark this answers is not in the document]
gtback: No, you could still pass in individual filters, which (if needed) could be put into a temporary FilterSet.

stix2/sources/__init__.py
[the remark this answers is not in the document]
gtback: Oh, I didn't realize it was a dict. I wonder if we shouldn't just make it a list. If we really need to support removing by ID, we can just loop through the list, but in reality I don't see this being used that often.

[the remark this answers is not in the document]
mbastian1135: okay, will modify

clenk:
From Travis, I think `test_data_sources.test_add_get_remove_filter()` is failing because of python 2 vs 3 differences. Maybe if we change the test to something like:

```
assert "Filter value type" in str(excinfo.value)
assert "not supported. The type must be a python immutable type or dictionary" in str(excinfo.value)
```

The only other thing is I think
`[stix_obj for stix_obj in apply_common_filters( ... )]`
can be simplified as
`list(apply_common_filters( ... ))`

But that's minor; I think it's good to go.

gtback:
I fixed the build failure and opened #59 to deal with the listcomps. One the builds pass, I'm going to merge this.

gtback:
Despite the coverage decreasing, I'm going to merge this. We have too many other things depending on a new release of python-stix2 to worry about the slight reduction (for now).

PR #60, opened 2017-10-03
Easier markings

### LINKED ISSUE #56: Improve Data Markings Ease-of-Use
Some ideas I had while writing documentation, for how to make data markings a little easier to use:

- [ ] Maybe `StatementMarking()` should return a `MarkingDefinition` object and not require the user to create both a `MarkingDefinition` and `StatementMarking` object. Would a user ever create a `StatementMarking` object without also creating a `MarkingDefinition`?
- [x] We should add a shortcut to be able to call `add_markings()` (and the other marking functions) on STIX objects. So for example
	```indicator.remove_markings('marking--id')```
	instead of
	```markings.remove_markings(indicator, 'marking--id')```
As we implement these, be explicit that they return *new* versions of the objects since they are immutable.
- [x] stix2/\_\_init\_\_.py should import the functions from stix2.markings so they're in the top-level 
namespace.
- [x] `add_markings()` and `remove_markings()` should support passing in a `MarkingDefinition` object, not just an ID string.

clenk:
Fixes #56.

### DIFF
--- a/stix2/markings/__init__.py
+++ b/stix2/markings/__init__.py
@@ -212,3 +212,16 @@ def is_marked(obj, marking=None, selectors=None, inherited=False, descendants=Fa
         result = result or object_markings.is_marked(obj, object_marks)
 
     return result
+
+
+class MarkingsMixin():
+    pass
+
+
+# Note that all of these methods will return a new object because of immutability
+MarkingsMixin.get_markings = get_markings
+MarkingsMixin.set_markings = set_markings
+MarkingsMixin.remove_markings = remove_markings
+MarkingsMixin.add_markings = add_markings
+MarkingsMixin.clear_markings = clear_markings
+MarkingsMixin.is_marked = is_marked
--- a/stix2/sdo.py
+++ b/stix2/sdo.py
@@ -6,6 +6,7 @@ import stix2
 
 from .base import _STIXBase
 from .common import ExternalReference, GranularMarking, KillChainPhase
+from .markings import MarkingsMixin
 from .observables import ObservableProperty
 from .properties import (BooleanProperty, IDProperty, IntegerProperty,
                          ListProperty, PatternProperty, ReferenceProperty,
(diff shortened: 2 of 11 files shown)


stix2/markings/__init__.py
```diff
@@ -212,3 +212,16 @@ def is_marked(obj, marking=None, selectors=None, inherited=False, descendants=Fa
         result = result or object_markings.is_marked(obj, object_marks)
 
     return result
+
+
+class MarkingsMixin():
+    pass
+
+
+# Note that all of these methods will return a new object because of immutability
```
gtback: Oooo, this is super-slick. Because those functions already take `obj`as their first argument, they can become methods on the Mixed-in objects (where `obj` is implicitly `self`).

stix2/sdo.py
```diff
@@ -6,14 +6,15 @@
...
+from .markings import MarkingsMixin
 from .observables import ObservableProperty
 from .properties import (BooleanProperty, IDProperty, IntegerProperty,
                          ListProperty, PatternProperty, ReferenceProperty,
                          StringProperty, TimestampProperty, TypeProperty)
 from .utils import NOW
 
 
-class AttackPattern(_STIXBase):
+class AttackPattern(_STIXBase, MarkingsMixin):
```
gtback: I think it's time to make an `STIXDomainObject` class that subclasses `_STIXBase` and `MarkingsMixin` and use that as the superclass for all SDOs.

stix2/sro.py
```diff
@@ -4,13 +4,14 @@
...
 from .common import ExternalReference, GranularMarking
+from .markings import MarkingsMixin
 from .properties import (BooleanProperty, IDProperty, IntegerProperty,
                          ListProperty, ReferenceProperty, StringProperty,
                          TimestampProperty, TypeProperty)
 from .utils import NOW
 
 
-class Relationship(_STIXBase):
+class Relationship(_STIXBase, MarkingsMixin):
```
gtback: Make a `STIXRelationshipObject` class to prevent adding the `MarkingsMixin` to both classes.

PR #61, opened 2017-10-03
frivolous list comprehensions converted to list()

### LINKED ISSUE #59: Make list comprehensions more idiomatic
We have several list comprehensions in the form:

```
[x for x in <iterable>]
```
These can be replaced with just `list(<iterable>)`

There are some more complex ones such as

```
[x for x in <iterable> if <some predicate on x>]
```

which can be replaced by `list(filter(lambda x: <predicate on x>, <iterable>))`

mbastian1135:
syntax tweaks to straightforward list comprehensions - fixes #59

### DIFF
--- a/stix2/sources/filesystem.py
+++ b/stix2/sources/filesystem.py
@@ -279,13 +279,11 @@ class FileSystemSource(DataSource):
                             # since ID is specified in one of filters, can evaluate against filename first without loading
                             stix_obj = json.load(open(os.path.join(root, file_)))["objects"][0]
                             # check against other filters, add if match
-                            matches = list(apply_common_filters([stix_obj], query))
-                            all_data.extend(matches)
+                            all_data.extend(apply_common_filters([stix_obj], query))
                     else:
                         # have to load into memory regardless to evaluate other filters
                         stix_obj = json.load(open(os.path.join(root, file_)))["objects"][0]
-                        matches = list(apply_common_filters([stix_obj], query))
-                        all_data.extend(matches)
+                        all_data.extend(apply_common_filters([stix_obj], query))
 
         all_data = deduplicate(all_data)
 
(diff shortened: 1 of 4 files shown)


stix2/sources/filesystem.py
```diff
@@ -279,12 +279,12 @@ def query(self, query=None, _composite_filters=None):
                             # since ID is specified in one of filters, can evaluate against filename first without loading
                             stix_obj = json.load(open(os.path.join(root, file_)))["objects"][0]
                             # check against other filters, add if match
-                            matches = [stix_obj_ for stix_obj_ in apply_common_filters([stix_obj], query)]
+                            matches = list(apply_common_filters([stix_obj], query))
                             all_data.extend(matches)
                     else:
```
chisholm: `apply_common_filters()` returns an iterable (a generator); `list.extend()` supports an iterable parameter.  Why not pass it in directly instead of making an intermediate list?  E.g. `all_data.extend(apply_common_filters([stix_obj], query))`?  This is related to my [other comment](https://github.com/oasis-open/cti-python-stix2/issues/59#issuecomment-333225791) about trying not to make copies unnecessarily.

stix2/sources/filesystem.py
```diff
@@ -279,12 +279,12 @@ def query(self, query=None, _composite_filters=None):
...
                         # have to load into memory regardless to evaluate other filters
                         stix_obj = json.load(open(os.path.join(root, file_)))["objects"][0]
-                        matches = [stix_obj_ for stix_obj_ in apply_common_filters([stix_obj], query)]
+                        matches = list(apply_common_filters([stix_obj], query))
                         all_data.extend(matches)
```
chisholm: Same issue here as above.

stix2/sources/filesystem.py
mbastian1135: No objection - I was not aware (or didnt remember) that extend() will take an iterable.

PR #63, opened 2017-10-04
Fix timestamp constant

rpiazza:
Minor fix...

Noticed that in pattern expressions timestamp constants are prefixed by 't'


rpiazza:
In general we don't have any tests for the constants, although some are created in the pattern_expressions tests.  Do you want a more comprehensive set of tests for all of the constant types?

gtback:
Eventually, yes, but not necessarily as part of this PR. I'm just trying to have the attitude of "test any code we make changes to", in the "leave things cleaner when you leave than when you arrived" spirit.

gtback:
Sorry I wasn't more explicit, @rpiazza . This is really all I wanted: 1414035e8dcbcee9ab5299b8801d242ef775c0ad

rpiazza:
Actually, I have been adding the examples from Part 5 of the spec for testing purposes, so I just added that one because it had a timestamp...

Issue #68, opened 2017-10-06
Required properties not enforced on certain SDOs


emmanvg:
Found the following SDOs not enforcing required properties:
- ObservedData
     - 'objects' is required
- Report
     - 'published' is required
     - 'object_refs' is required

One additional change:
- Indicator
     - 'labels' not in correct order.

emmanvg:
Missed:
- Mutex
     - 'name' is required

https://github.com/oasis-open/cti-python-stix2/blob/master/stix2/observables.py#L380

PR #72, opened 2017-10-06
Make Mutex 'name' required. Bunch of formatting changes, add test for custom extension

emmanvg:
Related to #68


clenk:
Looks nice! Should we also check that the custom extension's `properties` arg is not an empty dict (`{}`) in addition to checking that it's not `None`?

emmanvg:
That's right, we should also check is not empty. Good catch!

emmanvg:
Could you add it, as part of merging with master?

clenk:
Will do.

PR #78, opened 2017-10-11
STIX 2.1 Work in Progress

emmanvg:
The following additions have been made to support STIX 2.1:

- Added new STIX 2.1 objects (Location, Opinion and Note) and common properties (lang and confidence).
- Implemented all confidence scale transformations, both ways.
- Added LanguageContent object.

I have updated this branch with the latest in master as well.


emmanvg:
There are no tests yet for the new objects and methods.

gtback:
Looking good! I'm happy to merge this in now, or we can wait until later. Your call @emmanvg

emmanvg:
I think it would be good now. It will enable others to contribute directly in this branch rather than having it in my local fork. I will also open an issue for that (writing tests for new objects).

Issue #77, opened 2017-10-11
Missing properties and/or required


emmanvg:
Found a couple of Objects with missing properties or the property is not enforced:

- AutonomousSystem
     - number is `required`

- SocketExt
     - socket_descriptor is missing
     - socket_handle is missing

PR #81, opened 2017-10-11
Make AS 'number' required, add missing props to SocketExt.

emmanvg:
Related to #77


gtback:
This doesn't directly affect coverage, because it's technically *within* a single line that was already tested.

Issue #87, opened 2017-10-18
Unable to get() object with custom properties


FearonChris:
```
File "/usr/local/lib/python2.7/dist-packages/stix2/environment.py", line 110, in get
    return self.source.get(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/stix2/sources/__init__.py", line 262, in get
    data = ds.get(stix_id=stix_id, _composite_filters=all_filters)
  File "/usr/local/lib/python2.7/dist-packages/stix2/sources/__init__.py", line 262, in get
    data = ds.get(stix_id=stix_id, _composite_filters=all_filters)
  File "/usr/local/lib/python2.7/dist-packages/stix2/sources/filesystem.py", line 156, in get
    all_data = self.query(query=query, _composite_filters=_composite_filters)
  File "/usr/local/lib/python2.7/dist-packages/stix2/sources/filesystem.py", line 294, in query
    stix_objs = [parse(stix_obj_dict) for stix_obj_dict in all_data]
  File "/usr/local/lib/python2.7/dist-packages/stix2/core.py", line 94, in parse
    return obj_class(allow_custom=allow_custom, **obj)
  File "/usr/local/lib/python2.7/dist-packages/stix2/base.py", line 104, in __init__
    raise ExtraPropertiesError(cls, extra_kwargs)
stix2.exceptions.ExtraPropertiesError: Unexpected properties for Vulnerability:
```

clenk:
Thank you for this bug report, @FearonChris. The upcoming changes to fix #76 should fix this issue as well.

gtback:
Thanks, @FearonChris. We're working on this in conjunction with our work on #76.

FearonChris:
Apologies btw, I feel like i'm bombarding you guys! Thanks for your prompt responses and efforts on this project.

gtback:
No apologies needed, @FearonChris. Please keep the issues coming! It's good to see people using the libraries and exploring use cases (and code paths) that we haven't thought of (especially regarding custom properties, where it's harder for us to know the breadth of use cases).

Issue #76 led us to rethink how we are doing custom properties in the DataStore classes, which I think will result in a better overall design.

gtback:
Fixed by #88

Issue #85, opened 2017-10-12
Cannot serialize Bundle with object containing Custom Property


gtback:
To reproduce:

```python
 def test_custom_property_in_bundled_object():
        identity = stix2.Identity(
            name="John Smith",
            identity_class="individual",
            x_foo="bar",
            allow_custom=True,
        )
        bundle = stix2.Bundle(identity, allow_custom=True)

        assert bundle.objects[0].x_foo == "bar"
        assert '"x_foo": "bar"' in str(bundle)
```

Results in an error like this:

`TypeError: '<' not supported between instances of 'NoneType' and 'int'`

I believe this is because `stix2.utils.find_property_index` returns None for custom properties, which we should probably put at the end.

emmanvg:
I am unable to reproduce this error with the example you provided.

gtback:
I'm getting the error in Python 3, but not Python 2.

emmanvg:
Using source code in `76-filesystem-bundles` branch??

Well, that's weird.

emmanvg:
I tried running the tests in the `76-filesystem-bundles` branch. I get two failing errors, but these are related to datastores and it is completely different from what you are expressing on this issue. I tested with both Python 3 and 2, and I am still unable to reproduce this error. Do you have any local changes?

emmanvg:
Made the changes. It should be good now. The changes are present in `76-filesystem-bundles` branch.

clenk:
I can fix the remaining failing tests when I make the other changes for #76.

emmanvg:
Sounds good!

gtback:
We found this when working on #76, but I think the issue was present in the master branch. Seems like the fix was pretty easy, so it's fine that they are in the same branch.

PR #88, opened 2017-10-18
Filesystem Bundles and custom content

### LINKED ISSUE #76: Unable to add() Bundle.
example usage:

```
bundle      = Bundle(*bArgs, allow_custom=True) 
store.add(bundle)
```

```
  File "stix.py", line 166, in <module>
    store.add(bundle)
  File "/usr/local/lib/python2.7/dist-packages/stix2/environment.py", line 147, in add
    return self.sink.add(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/stix2/sources/filesystem.py", line 86, in add
    _check_path_and_write(self._stix_dir, stix_data)
  File "/usr/local/lib/python2.7/dist-packages/stix2/sources/filesystem.py", line 82, in _check_path_and_write
    f.write(str(Bundle(stix_obj)))
  File "/usr/local/lib/python2.7/dist-packages/stix2/core.py", line 62, in __init__
    super(Bundle, self).__init__(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/stix2/base.py", line 121, in __init__
  File "/usr/local/lib/python2.7/dist-packages/stix2/base.py", line 57, in _check_property
    raise InvalidValueError(self.__class__, prop_name, reason=str(exc))
stix2.exceptions.InvalidValueError: Invalid value for Bundle 'objects': This property may not contain a Bundle object
```

FYI i'm using the patched code submitted in #73

clenk:
- Filesystem now saves objects in file without wrapping them in bundles, but can parse objects saved in bundles in files.
- FileSystem and Memory tests are separated out into their own files.
- Custom properties, objects, etc. can be added to and retrieved from FileSystem and Memory datastores.
- Custom properties are returned by object.

Fixes #76, #85, and #87.


stix2/sources/__init__.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, source=None, sink=None):
         self.source = source
         self.sink = sink
 
-    def get(self, stix_id):
+    def get(self, stix_id, allow_custom=False):
```
gtback: Can we add the new argument to the docstring? (same in several locations below)

clenk:
Don't know why Travis failed yesterday, but I reran the job and it passed.

stix2/sources/memory.py
```diff
@@ -283,7 +283,7 @@ def query(self, query=None, _composite_filters=None, allow_custom=False):
             if not isinstance(query, list):
                 # make sure dont make set from a Filter object,
                 # need to make a set from a list of Filter objects (even if just one Filter)
-                query = list(query)
+                query = [query]
```
gtback: Can we replace this entire if block with just `query = list(query)`? I agree it's unnecessary with the isinstance check. In fact, we could combine with the below `query = set(list(query))` (which is a fairly common idiom I've seen). Some good tests would help ensure this part behaves as expected. In fact, we could probably pull out a function `def get_filter_set(query)` or put that as a method on the base `DataSource class`. 

We don't need to worry about this for this PR, but maybe make an issue for it?

stix2/sources/filesystem.py
```diff
@@ -78,45 +83,41 @@ def _check_path_and_write(stix_dir, stix_obj):
                 os.makedirs(os.path.dirname(path))
 
             with open(path, "w") as f:
-                # Bundle() can take dict or STIX obj as argument
-                f.write(str(Bundle(stix_obj)))
+                f.write(str(stix_obj))
```
gtback: Does this mean the behavior is now changed to not save a Bundle in each file? Can we add a parameter to the FileSystemSink that retains the current behavior (by essentially doing `if self.bundlify: stix_obj = Bundle(stix_obj)`?) It shouldn't be the default, and we don't need to worry about passing the parameter through via higher levels, but we should at least allow an option for the current behavior. 

Also, `_check_path_and_write` doesn't need to be a nested function, and without the bundlify option, it doesn't even need to be inside this class.

stix2/test/test_bundle.py
```diff
@@ -158,3 +158,11 @@ def test_parse_unknown_type():
...
         stix2.parse(unknown)
     assert str(excinfo.value) == "Can't parse unknown object type 'other'! For custom types, use the CustomObject decorator."
+
+
+def test_stix_object_property():
+    prop = stix2.core.STIXObjectProperty()
+
+    identity = stix2.Identity(name="test", identity_class="individual")
+    assert prop.clean(identity) == identity
+    assert prop.clean(identity) is identity
```
gtback: We don't need both `==` and `is` checks.

stix2/sources/memory.py
chisholm: Correct me if I'm wrong, but the point seems to be to allow the method to accept either a single filter, or a list of filters.  A filter is itself an iterable (it's a tuple), so `list(query)`, if `query` is a single filter, would result in a list like `["prop", "=", "foo"]`, not what is intended.

Also, `set(list(query))` looks to me like it creates an unnecessary intermediate list.  Once you've ensured `query` is an iterable of filters, `set(query)` should create the filter set.

clenk: @chisholm has the right of it. I got an error in the tests with `list(query)` but not `[query]`.

stix2/test/test_bundle.py
chisholm: In theory, this could be appropriate, if you wanted to test an `__eq__` override and make sure it's consistent with `is`.  But the stix objs don't have that override do they?  So, perhaps unnecessary in this case.  An `__eq__` could be defined somewhere perhaps, which checks the `id` and `modified` properties of stix objects.  Then a test like this could make sense I think.

stix2/sources/filesystem.py
clenk: Yes, it won't save as a bundle. I'll add that option just to the FileSystemSink.

gtback: Yep, adding to the Store makes sense.

stix2/sources/memory.py
gtback: Yes, you're right. Nothing to see here... move along 😁

stix2/test/test_bundle.py
gtback: I'm pretty sure `is` ensures the values point to the same item in memory. I believe we get `__eq__` for free with `collections.abc.Mapping`, but it's not as useful for tests since you would need to ensure even the auto-generated properties like `id` and `created` are the same.

Each tests has its purpose, but in this case I'm not sure which is more important, that `prop.clean` returns the same object, or just an equivalent object.

clenk: This test is making sure that instances of our STIX classes get added to Bundles without being re-parsed/dictified: https://github.com/oasis-open/cti-python-stix2/pull/88/commits/374539a6cc7fe0f380ab598ed39d791967119afb#diff-0b92af30cf45a78c5cf9a35dd394cfeaR22

Thus I think the `is` check is the correct one in this case.

Issue #89, opened 2017-10-20
Add functions to get related objects


clenk:
In preparation for the Workbench layer, the Environment API should have some functions to explore/retrieve/dereference related objects. This would include something like `created_by()` (for `created_by_ref`) and `relationships()`. `relationships()` might have parameters like `source_only`, `target_only`, and `relationship_type`.

We will need to modify Filters to allow filtering on non-common properties first, however, as we cannot currently filter on `source_ref` and `target_ref`.

Steps:
- [x] `created_by()`/`creator_of()`
- [x] filters
- [ ] `relationships()`
- [ ] `related_to()`

PR #90, opened 2017-10-20
Add created_by() function to Environment

clenk:
See also #89.

### DIFF
--- a/stix2/environment.py
+++ b/stix2/environment.py
@@ -152,3 +152,17 @@ class Environment(object):
     def parse(self, *args, **kwargs):
         return _parse(*args, **kwargs)
     parse.__doc__ = _parse.__doc__
+
+    def created_by(self, obj):
+        """Retrieve the Identity refered to by the object's `created_by_ref`.
+
+        Args:
+            obj: The STIX object whose `created_by_ref` property will be looked
+                up.
+
+        Returns:
+            The STIX object's creator.
+        """
+
+        creator_id = obj.get('created_by_ref', '')
+        return self.get(creator_id)
--- a/stix2/environment.py
+++ b/stix2/environment.py
@@ -153,7 +153,7 @@ class Environment(object):
         return _parse(*args, **kwargs)
     parse.__doc__ = _parse.__doc__
 
-    def created_by(self, obj):
+    def creator_of(self, obj):
         """Retrieve the Identity refered to by the object's `created_by_ref`.
 
         Args:
@@ -161,8 +161,13 @@ class Environment(object):
                 up.
 
         Returns:
-            The STIX object's creator.
-        """
+            The STIX object's creator, or
+            None, if the object contains no `created_by_ref` property or the
+                object's creator cannot be found.
 
+        """
         creator_id = obj.get('created_by_ref', '')
-        return self.get(creator_id)
+        if creator_id:
+            return self.get(creator_id)
+        else:
+            return None
--- a/stix2/sources/memory.py
+++ b/stix2/sources/memory.py
@@ -215,10 +215,14 @@ class MemorySource(DataSource):
 
         all_data = self.query(query=query, _composite_filters=_composite_filters)
 
-        # reduce to most recent version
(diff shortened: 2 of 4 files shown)


stix2/environment.py
```diff
@@ -152,3 +152,17 @@ def add(self, *args, **kwargs):
     def parse(self, *args, **kwargs):
         return _parse(*args, **kwargs)
     parse.__doc__ = _parse.__doc__
+
+    def created_by(self, obj):
```
gtback: I'm trying to think about the best name for this function. I'm leaning towards `creator_of()` or `get_creator_of()`, since the idea behind the Workbench API layer is that all of the environment methods will be exposed as module-level functions on the "implicit `Environment`" of the Workbench. The `get_` prefix is (to me) an indicator that the operation is potentially expensive (depending on the type of DataSource), but I don't know if we want a ton of `get_*` functions in the Workbench API.

stix2/environment.py
```diff
@@ -152,3 +152,17 @@ def add(self, *args, **kwargs):
...
+
+        Args:
+            obj: The STIX object whose `created_by_ref` property will be looked
+                up.
+
+        Returns:
+            The STIX object's creator.
+        """
+
+        creator_id = obj.get('created_by_ref', '')
```
gtback: If there is no `created_by_ref`, we should probably `return None` rather than calling the line below.

stix2/environment.py
```diff
@@ -152,3 +152,17 @@ def add(self, *args, **kwargs):
...
+        return self.get(creator_id)
```
gtback: What should we do if the object has a `created_by_ref` but it can't be found via the configured DataSource? DataSource objects currently return None in this case, but I'm starting to wonder if that's the best approach; even if it is, I'm not sure if we want to make some distinction in this function.

stix2/test/test_environment.py
```diff
@@ -184,3 +184,25 @@ def test_parse_malware():
...
+
+def test_created_by():
+    identity = stix2.Identity(**IDENTITY_KWARGS)
+    factory = stix2.ObjectFactory(created_by_ref=identity.id)
+    env = stix2.Environment(store=stix2.MemoryStore(), factory=factory)
+    env.add(identity)
+
+    ind = env.create(stix2.Indicator, **INDICATOR_KWARGS)
+    creator = env.created_by(ind)
+    assert creator.id == identity.id
```
gtback: The actual objects should be equal, not just their ID's, correct? It might help to be clear that the objects themselves are equal.

stix2/test/test_environment.py
```diff
@@ -184,3 +184,25 @@ def test_parse_malware():
     assert mal.modified == FAKE_TIME
     assert mal.labels == ['ransomware']
     assert mal.name == "Cryptolocker"
+
```
gtback: Add a test for when there is a configured data source, but that sources does not have the desired object.

stix2/environment.py
clenk: I thought about this, too. My reason for going with this was an idea that in the workbench layer you could do `myobj.created_by()` to get the creator object. But since that would be a wrapper around this function I guess it doesn't need to be named the same (i.e. `myobj.created_by()` could call `creator_of(myobj)` under the hood).

I'll rename it.

clenk:
I went with just returning `None` for now, but we can revisit if we want to return an empty list or raise some sort of `NotFound` exception.

stix2/sources/memory.py
```diff
@@ -215,10 +215,14 @@ def get(self, stix_id, _composite_filters=None):
 
         all_data = self.query(query=query, _composite_filters=_composite_filters)
 
-        # reduce to most recent version
-        stix_obj = sorted(all_data, key=lambda k: k['modified'])[0]
+        if all_data:
+            print(all_data)
```
gtback: Do we need this print statement?

gtback:
I'm ok with using `None` in both cases and not distinguishing between "no `created_by_ref`" and "lookup unsuccessful". For now, if someone cares about the distinction, they can check the corresponding property themselves.

Issue #66, opened 2017-10-06
TAXIICollectionSource.get()/all_versions() - check id of returned objects


mbastian1135:
Add simple sanity check for the TAXIICollectionSource.get()/all_versions() calls, where the id of the STIX objects returned from the TAXII endpoint are checked to match

mbastian1135:
Have this implemented locally. Was going to wait(?) for Brets external server to work.; or can just push.

gtback:
Go ahead and push/make a PR now. I'm not sure the best way to test (and if there's a good way without an external server), but we should at least start thinking about it.

rpiazza:
Ignoring Travis, are you using the cti_taxii_server to test in your own environment?

mbastian1135:
I was not but I can.

gtback:
Thanks, @rpiazza . Using medallion for local testing is 100% the way to go; testing via Travis will be a bit harder. We may need to set up some mock responses.

Issue #64, opened 2017-10-05
Add tests for TAXIICollection Data Stores, Sources, Sinks


gtback:
From #58

PR #92, opened 2017-10-25
66 - checking STIX obj ID from TAXII server

mbastian1135:
issue #66


clenk:
Do we need to check the Id of returned objects twice? Seems like we're doing it both on line 129 and line 163.

gtback:
@clenk : I think one is in `get` and the other is in `all_versions`.

docs/guide/taxii.ipynb
```diff
@@ -2406,7 +159,6 @@
     "indicators = tc_source.query([f1])\n",
```
clenk: This is minor, but at least when I tested with medallion's test data, this didn't return multiple objects, but using a filter of `!=` did.

docs/guide/taxii.ipynb
```diff
@@ -71,2326 +71,79 @@
...
-      "    \"valid_from\": \"2017-10-02T20:40:44Z\",\n",
-      "    \"kill_chain_phases\": [\n",
-      "        {\n",
-      "            \"kill_chain_name\": \"lockheed-martin-cyber-kill-chain\",\n",
-      "            \"phase_name\": \"delivery\"\n",
-      "        }\n",
-      "    ]\n",
-      "}\n",
-      "{\n",
-
```
clenk: Not a big deal either way, but maybe change the password to the one used in medallion's readme and test config (`Password0`).

stix2/sources/taxii.py
```diff
@@ -1,8 +1,9 @@
 """
-Python STIX 2.0 TAXII Source/Sink
-
-TODO:
-    Test everything
+Python STIX 2.x
```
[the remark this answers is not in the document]
mbastian1135: Yea this is old...before I knew how to do proper docs.

stix2/sources/taxii.py
```diff
@@ -121,11 +122,13 @@ def get(self, stix_id, _composite_filters=None):
         # as directly retrieveing a STIX object by ID
         stix_objs = self.collection.get_object(stix_id)["objects"]
 
-        stix_obj = list(apply_common_filters(stix_objs, query))
+        stix_obj = [apply_common_filters(stix_objs, query)]
```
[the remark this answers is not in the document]
mbastian1135: Agree.

docs/guide/taxii.ipynb
mbastian1135: That was a local hack to make sure I was using the right instance of medallion. So agree.

docs/guide/taxii.ipynb
mbastian1135: No it doesnt. I was just pretending that it did (or would have).  Originally, when querying Brets server, we would get 126 indicators in return. And since he only had indicators, the example was just pretending that there was a mix of objects but we were only getting indicators back, when in reality you only got all the indicators every time (bc was broken).

The example is still correct, it just might look off as its iterating over the list and there is only one.

docs/guide/taxii.ipynb
mbastian1135: and noted on the #64 issue

stix2/sources/taxii.py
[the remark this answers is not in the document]
mbastian1135: Agree.

stix2/sources/taxii.py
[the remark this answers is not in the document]
mbastian1135: So [] will not run the generator through to make a list? Interesting

clenk:
Thank you, @mbastian1135!

PR #93, opened 2017-10-26
Update stix2 Package Structure 

### LINKED ISSUE #91: Update package structure
Per conversation in #79 we should start by updating the current package structure for STIX 2.0

emmanvg:
Closes #91

### DIFF
--- a/stix2/__init__.py
+++ b/stix2/__init__.py
@@ -19,27 +19,11 @@
 
 # flake8: noqa
 
-from . import exceptions
-from .common import (TLP_AMBER, TLP_GREEN, TLP_RED, TLP_WHITE, CustomMarking,
-                     ExternalReference, GranularMarking, KillChainPhase,
-                     MarkingDefinition, StatementMarking, TLPMarking)
-from .core import Bundle, _register_type, parse
+from . import exceptions, v20
+from .core import Bundle, _collect_stix2_obj_maps, _register_type, parse
 from .environment import Environment, ObjectFactory
 from .markings import (add_markings, clear_markings, get_markings, is_marked,
                        remove_markings, set_markings)
-from .observables import (URL, AlternateDataStream, ArchiveExt, Artifact,
-                          AutonomousSystem, CustomExtension, CustomObservable,
-                          Directory, DomainName, EmailAddress, EmailMessage,
-                          EmailMIMEComponent, File, HTTPRequestExt, ICMPExt,
-                          IPv4Address, IPv6Address, MACAddress, Mutex,
-                          NetworkTraffic, NTFSExt, PDFExt, Process,
-                          RasterImageExt, SocketExt, Software, TCPExt,
-                          UNIXAccountExt, UserAccount, WindowsPEBinaryExt,
-                          WindowsPEOptionalHeaderType, WindowsPESection,
-                          WindowsProcessExt, WindowsRegistryKey,
-                          WindowsRegistryValueType, WindowsServiceExt,
-                          X509Certificate, X509V3ExtenstionsType,
-                          parse_observable)
 from .patterns import (AndBooleanExpression, AndObservationExpression,
                        BasicObjectPathComponent, EqualityComparisonExpression,
                        FloatConstant, FollowedByObservationExpression,
@@ -58,9 +42,6 @@ from .patterns import (AndBooleanExpression, AndObservationExpression,
                        ReferenceObjectPathComponent, RepeatQualifier,
                        StartStopQualifier, StringConstant, TimestampConstant,
                        WithinQualifier)
-from .sdo import (AttackPattern, Campaign, CourseOfAction, CustomObject,
-                  Identity, Indicator, IntrusionSet, Malware, ObservedData,
-                  Report, ThreatActor, Tool, Vulnerability)
 from .sources import CompositeDataSource
 from .sources.filesystem import (FileSystemSink, FileSystemSource,
                                  FileSystemStore)
@@ -68,6 +49,8 @@ from .sources.filters import Filter
 from .sources.memory import MemorySink, MemorySource, MemoryStore
 from .sources.taxii import (TAXIICollectionSink, TAXIICollectionSource,
                             TAXIICollectionStore)
-from .sro import Relationship, Sighting
 from .utils import get_dict, new_version, revoke
+from .v20 import *  # This import should always be the latest STIX 2.X version
 from .version import __version__
+
+_collect_stix2_obj_maps()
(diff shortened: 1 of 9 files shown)


emmanvg:
I am going to write something for documentation. We should make it clear that using just doing `import stix2` will change your behavior with every major release.

I could also modify the settings in the packaging to remove the tests from the packages distributed officially through PyPI.

stix2/__init__.py
```diff
@@ -19,27 +19,11 @@
 
 # flake8: noqa
 
-from . import exceptions
-from .common import (TLP_AMBER, TLP_GREEN, TLP_RED, TLP_WHITE, CustomMarking,
-                     ExternalReference, GranularMarking, KillChainPhase,
-                     MarkingDefinition, StatementMarking, TLPMarking)
-from .core import Bundle, _register_type, parse
+from . import exceptions, v20
```
gtback: I'm not sure this line is needed. `stix2.exceptions` and `stix2.v20` should both be importable even without this line.

stix2/v20/sdo.py
```diff
@@ -358,7 +358,7 @@ def __init__(self, **kwargs):
                         return
                     raise e
 
-        stix2._register_type(_Custom)
+        stix2._register_type(_Custom, version="2.0")
```
gtback: Part of me wonders if custom objects should be version-specific. Do we envision things about how custom objects work changing between 2.0 and 2.1?

stix2/__init__.py
emmanvg: You are correct. We can remove these.

stix2/v20/sdo.py
emmanvg: What I had in mind is to explicitly register Custom objects for each STIX version in order to be registered. I think the developers need to be aware that properties and requirements change between version. So their objects might behave differently across STIX 2.X versions as well.

If we register Custom objects for all versions at once, for example new properties might be automatically added, their behavior might change based on changing requirements.

Also, all dictionary mappings to track internal and custom objects need to updated.

If you think a register for all versions is ok. We could have a higher level constructor that appropriately delegates the call to the correct Custom decorator and register method.

emmanvg:
Also, using `python setup.py sdist' I managed to remove the `test` dir from the package, but I cannot get it to work with wheel `python setup.py bdist_wheel`.... I will push what I have.

stix2/v20/sdo.py
clenk: Yeah, the common properties may change between versions. Also, a user might create a custom object with an `__init__()` that makes use of one of the new common properties. What do we do if that gets registered with an older version of the spec that doesn't have that property?

That's one reason to go for version-specific registering.

packet-rat: @gtback :  I think one thing we can absolutely predict are substantive changes in point releases as all of the objects we "punted on" get defined/refined in 2.1, 2.2, ... 

So the decisions here on strategy should presume fairly disruptive changes to these libraries.  Most of of out here (self included) are not going to understand all of the underlying nuances of _ .vs. __ .vs. ___ et al.

So to the extent you can maintain an abstraction layer to protect "black box" users down the road it will be much appreciated.

docs/ts_support.rst
```diff
@@ -0,0 +1,131 @@
+How imports will work
```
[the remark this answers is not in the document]
gtback: Agree! This documentation is excellent! it's not really code that should get executed, as all the actual details other than the module namespaces are elided. I'm fine either way, though

README.rst
```diff
@@ -62,6 +62,15 @@ To parse a STIX JSON string into a Python STIX object, use ``parse()``:
 
 For more in-depth documentation, please see `https://stix2.readthedocs.io/ <https://stix2.readthedocs.io/>`__.
 
+STIX 2.X Technical Specification Support
+----------------------------------------
+
+The `stix2` Python library is built to support multiple versions of the STIX
```
gtback: May want to put an explicit statement of "This version of python-stix2 supports STIX 2.0 by default", and then change this statement as new versions of python-stix2 supporting newer versions of STIX" are released.

docs/ts_support.rst
```diff
@@ -0,0 +1,131 @@
...
+CustomObjects, CustomObservable, CustomMarking and CustomExtensions must be
+registered explicitly by STIX version. This is a design decision since properties
+or requirements may chance as the STIX Technical Specification advances.
+
+You can perform this by,
+
+.. code:: python
+
+    # Make my custom observable available in STIX 2.0
+    @stix2.v20.observables.CustomObservable('x-new-object-type',
```
gtback: We should hide the `observables` element of this module path, and instead ensure that `@stix2.v20.CustomObservable` works. All of the internals of the `stix2.v2<x>` modules should be flattened.

stix2/core.py
```diff
@@ -62,37 +59,30 @@ def __init__(self, *args, **kwargs):
...
+        version (str): Which STIX2 version to use. (e.g. "2.0", "2.1"). If
+            None, use latest version.
 
     Returns:
         An instantiated Python STIX object.
 
     """
+    if not version:
+        # Use latest version
+        OBJ_MAP = STIX2_OBJ_MAPS[sorted(STIX2_OBJ_MAPS.keys())[-1]]
```
gtback: Would it help to have a global `stix2.DEFAULT_VERSION` variable that makes it easier for code that uses python-stix2 to check what version of STIX is supported by default?

stix2/v20/sdo.py
gtback: Thanks, @packet-rat. I agree with the conversation here that it's better to be explicit and require that custom types be registered explicitly with each version of STIX. This would also allow people to support (for example) the 2.1-only Location SDO in STIX 2.0 as a Custom Object, without conflicting implementations (I think).

My original question was whether something might be a "valid" 2.0 Custom Object, but no longer be valid in STIX 2.1 (or vice versa). As far as I can tell, there won't be any changes in 2.1, but who knows in 2.2 or later.

emmanvg:
The latest commits should address all of the aforementioned concerns.

stix2/core.py
```diff
@@ -21,8 +21,8 @@ def __init__(self, allow_custom=False):
     def clean(self, value):
         # Any STIX Object (SDO, SRO, or Marking Definition) can be added to
         # a bundle with no further checks.
-        if isinstance(value, (STIXDomainObject, STIXRelationshipObject,
-                              MarkingDefinition)):
+        if any(x in ('STIXDomainObject', 'STIXRelationshipObject', 'MarkingDefinition')
```
emmanvg: Given that STIXDomainObject and STIXRelationshipObject are just wrappers, can we move these to a higher level and prevent the current approach? Possibly create a new one for MarkingDefinition (perhaps STIXObject?)

gtback: Yes, we can create some base classes at the higher level. These are really just wrappers for groupings of STIX Types. The specs seem somewhat ambiguous whether MarkingDefinition (and Bundle) are STIX objects or not, so I'd rather keep those in groups by themselves.

PR #95, opened 2017-10-30
Change filters to allow filtering all properties

clenk:
(Not just common properties).

See #89.

We may want to revisit filters later so you can have a filter that matches if _any_ of the `objects` in `observed-data` match instead of having to specify the specific one. But that seems more complicated than is needed for dereferencing relationships for #89 so I left it for now.

### DIFF
--- a/stix2/sources/filters.py
+++ b/stix2/sources/filters.py
@@ -4,7 +4,6 @@ Filters for Python STIX 2.0 DataSources, DataSinks, DataStores
 """
 
 import collections
-import types
 
 # Currently, only STIX 2.0 common SDO fields (that are not complex objects)
 # are supported for filtering on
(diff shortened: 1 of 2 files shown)


stix2/sources/filters.py
```diff
@@ -155,7 +116,53 @@ def apply_common_filters(stix_objs, query):
...
+        # (i.e. did not make it through the filter)
+        return False
+
+    if "." in filter_.field:
+        # Check embedded properties, from e.g. granular_markings or external_references
+        sub_field = filter_.field.split(".", 1)[1]
+        sub_filter = filter_._replace(field=sub_field)
+        if isinstance(stix_obj[field], list):
+            for elem in stix_obj[field]:
+                r = _check_filter(sub_filter, elem)
```
gtback: Since `_check_filter` returns a boolean, this might be clearer as:
```python
if _check_filter(sub_filter, elem) is True:
    return True
```

stix2/sources/filters.py
```diff
@@ -155,7 +116,53 @@ def apply_common_filters(stix_objs, query):
...
+    elif isinstance(stix_obj[field], list):
+        # Check each item in list property to see if it matches
+        for elem in stix_obj[field]:
+            r = _all_filter(filter_, elem)
+            if r:
+                return r
+        return False
+    else:
+        # Check if property matches
+        return _all_filter(filter_, stix_obj[field])
```
gtback: maybe rename `_all_filter` to something else? Consider making it a method on the Filter class, since it touches two of the three data elements?

clenk: I like this idea. Maybe call it `apply` or `apply_to_property`?

Which reminds me, I need to make a pass through this and make sure we call them `property` instead of `field`.

stix2/sources/filters.py
clenk: I was thinking we couldn't check if `is True` because `_all_filter` returns -1 if the filter operation is not supported, but I think it would be better to have `_all_filter` raise an error instead of returning the "-1" special value.

PR #97, opened 2017-11-01
Test MemoryStore saving/loading to/from file

### LINKED ISSUE #65: Add tests for Memory Data Stores, Sources, Sinks
From #58

clenk:
Python 3 dict.values() returns a view, not a list. See https://stackoverflow.com/a/17431716/

Brings memory store code coverage to 100%.

Fixes #65.

### DIFF
--- a/stix2/sources/memory.py
+++ b/stix2/sources/memory.py
@@ -164,7 +164,7 @@ class MemorySink(DataSink):
         if not os.path.exists(os.path.dirname(file_path)):
             os.makedirs(os.path.dirname(file_path))
         with open(file_path, "w") as f:
-            f.write(str(Bundle(self._data.values(), allow_custom=allow_custom)))
+            f.write(str(Bundle(list(self._data.values()), allow_custom=allow_custom)))
     save_to_file.__doc__ = MemoryStore.save_to_file.__doc__
 
 
--- a/stix2/test/test_memory.py
+++ b/stix2/test/test_memory.py
@@ -1,3 +1,6 @@
+import os
+import shutil
+
 import pytest
 
 from stix2 import (Bundle, Campaign, CustomObject, Filter, MemorySource,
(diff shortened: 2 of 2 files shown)


stix2/sources/memory.py
```diff
@@ -164,7 +164,7 @@ def save_to_file(self, file_path, allow_custom=False):
         if not os.path.exists(os.path.dirname(file_path)):
             os.makedirs(os.path.dirname(file_path))
         with open(file_path, "w") as f:
-            f.write(str(Bundle(self._data.values(), allow_custom=allow_custom)))
+            f.write(str(Bundle(list(self._data.values()), allow_custom=allow_custom)))
```
gtback: Would it make sense for the Bundle constructor to consume any iterator that gets passed to it, rather than forcing callers to do this?

chisholm: I certainly like sticking to the most general interfaces you can get away with.  I think you mean you'd like the Bundle ctor to support any _iterable_.  Maybe it could support iterables for any of the positional parameters, and concatenate everything together?

clenk: I don't think we could do this for all iterables - for example, dictionaries are iterable but we don't want to call list() on a dictionary that's supposed to be a stix object being added to the bundle.

We only found this bug because the MemoryStore saves STIX content in a dictionary where the STIX object id is the key and the STIX object is the value. Could we add a check to the Bundle constructor like:

```python
elif isinstance(args[0], dict_values):
    kwargs['objects'] = list(args[0]) + list(args[1:]) + kwargs.get('objects', [])
```

chisholm: Hmm... yeah, that's a typical caveat, e.g. even strings are iterable.  Several times I've written a function that essentially checks iterable-but-not-a-string, to weed out the "funny" iterables.  I guess something similar could be done here.  E.g. iterable but not a dict or _STIXBase (which are the only two funny iterables that might be passed to this ctor?).

To me, just adding a special case for dict_values still doesn't seem as generic as it could be...

gtback: We can look into this more in the future, but I think there are enough exceptions (string, dict, _STIXBase) that it's not worth doing right now.

PR #100, opened 2017-11-03
Performance, Design and other fixes

### LINKED ISSUE #96: Something in the Bundle() function is causing a serious delay for building large STIX bundles.
Something that was changed in the Bundle() function between v0.2 and v0.3 is introducing large delays in creating a bundle for very large bundles (greater than a few hundlred STIX objects).

In v0.2, building a bundle containing 1000 STIX objects would take a few seconds.  In v0.3, it's taking 10+ minutes.  It might be related to the Bundle() function trying to convert things to an OrderedDict.

### LINKED ISSUE #98: Custom properties not added when using allow_custom=True

### LINKED ISSUE #99: Update DataStores, Sources, Sinks
While working #98 I noticed even further problems with how the stores work. In addition to the fixes, I will adding:
- Proper ABC for DataStore, DataStore, and DataSink.
- The version positional argument was missing in a lot of places.

emmanvg:
The major changes include:
- A new ``serialize()`` method to add flexibility for performance. You can pass the same arguments you would for a ``json.dumps()`` call. For example, `sort_keys=True`, `indent=4`, etc.
- Created a `CustomProperty` class, which developers can extend from to create their custom properties. It is wrapper, but it helps to signal the usage of custom properties. This fixes the problem when an object is parsed with `allow_custom=True` but the _properties dictionary would not get updated.
- ABC for `DataStore`, `DataSink` and `DataSource`. I think the changes made are good for proper concrete implementations of these ABCs. For example, DataStore was never abstract...

closes #99
closes #98
closes #96


clenk:
We talked about this offline, but just to document it: We also need to make sure custom properties' ordering works if they are passed to the constructor via a `custom_properties` dictionary instead of `allow_custom`. Like the 2nd example on https://stix2.readthedocs.io/en/latest/guide/custom.html#Custom-Properties.

mbastian1135:
Why are we making the classes abstract now? (bc thats how I had them originally but then we decided not to have ABC's)

gtback:
@clenk: Is there a reason you think they would be handled differently? Everything gets collapsed into the same dictionary, whether they came from `custom_properties` dict or a standard kwarg (when `allow_custom=True`). It's interesting (and probably not desirable) that the custom properties come first in both examples on the Jupyter notebook you linked to.

stix2/base.py
```diff
@@ -109,6 +102,13 @@ def __init__(self, allow_custom=False, **kwargs):
             extra_kwargs = list(set(kwargs) - set(cls._properties))
             if extra_kwargs:
                 raise ExtraPropertiesError(cls, extra_kwargs)
+        else:
+            from .properties import CustomProperty
+
+            # The custom properties will get added to the bottom.
+            # Matched with a CustomProperty.
+            extra_kwargs = list(set(kwargs) - set(cls._properties))
+            self._properties.update([(x, CustomProperty()) for x in extra_kwargs])
```
gtback: I'm not sure we want to do this. It updates the class instance of `_properties`, meaning that any custom properties on the first instantiation of an SDO are allowed on any future instance, even if allow_custom is false for the future ones.

It's perhaps likely that people want to use the same custom properties on every instance of the same SDO that they create, but I'm not sure how I feel about the way this is done.

emmanvg: That is a good point! So, could we (possible approaches):
- Make _properties an instance variable?
- Handle custom in a completely different structure?
- Go back with the old method (required checking the _inner dict)?

emmanvg: I'll go back to the #85 fix. That was working fine.

gtback:
I'd like to figure out better ways to track the version history of the Jupyter notebooks. I noticed that a couple switched from Python 2 to Python 3 (which is OK, but I just want to make sure what we do is deliberate). @clenk , @mbastian1135 any ideas?

stix2/sources/__init__.py
```diff
@@ -11,104 +11,106 @@
...
     """
     def __init__(self, source=None, sink=None):
+        super(DataStore, self).__init__()
         self.id = make_id()
         self.source = source
         self.sink = sink
 
-    def get(self, stix_id, allow_custom=False):
+    @abstractmethod
+    def get(self, stix_id):  # pragma: no cover
```
gtback: Rather than making these abstract methods, can we just define them similar to `get(self, *args, **kwargs)`, and pass them on the source? I'd rather not require subclasses to override these.

clenk:
@gtback The notebooks will always switch Python version when someone re-executes cells in them unless we all use the exact same version of Python. I think we could use a git filter to ignore the lines about python version. See https://stackoverflow.com/q/16244969/ We could have git ignore the `execution_count` lines as well.

If you're referring to tracking changes in the content of output cells instead, there are some ways to handle that too: http://nbsphinx.readthedocs.io/en/0.2.16/usage.html#Using-Notebooks-with-Git.

gtback:
@clenk: I just want to make sure we don't accidentally introduce extraneous diffs in the Notebook JSON files (output, Python version, execution count, etc.). That link to the nbsphinx output is extremely useful, but do we really want to not pre-run any of the cells? Will that even work on ReadTheDocs? (sorry, more questions than answers, I know)

stix2/sources/__init__.py
```diff
@@ -11,104 +11,106 @@
...
-            allow_custom (bool): whether to allow custom objects/properties or
-                not. Default: False.
+
         """
-        return self.sink.add(stix_objs, allow_custom=allow_custom)
+        return NotImplementedError()
 
 
-class DataSink(object):
+class DataSink(with_metaclass(ABCMeta)):
```
gtback: Having DataSink and DataSource as abstract classes is OK. I just think DataStore can provide a lot of default functionality (delegating to `source` and `sink` attributes.

stix2/sources/filesystem.py
```diff
@@ -35,6 +35,85 @@ def __init__(self, stix_dir, bundlify=False):
         self.source = FileSystemSource(stix_dir=stix_dir)
         self.sink = FileSystemSink(stix_dir=stix_dir, bundlify=bundlify)
 
+    def get(self, stix_id, allow_custom=False, version=None, _composite_filters=None):
```
gtback: None of this is specific to the FileSystem, I don't think. I'm hoping we can suitably abstract it with `*args` and `**kwargs` in the base class.

clenk:
As the Jupyter notebook versioning discussion seems tangential to this PR, I've moved it to #101.

emmanvg:
Hopefully I have addressed the concerns with the DataStores. It looks a lot cleaner now. Note: I did not change anything related to Jupyter Notebook since there is a new issue open. If nothing else needs to be fixed i'd say its ready to be merged! 👍

Issue #109, opened 2017-11-15
Allow CustomMarking without __init__()


emmanvg:
The CustomMarking decorator is missing the try/catch block CustomObject, and the other custom decorators have for empty inits.

PR #110, opened 2017-11-15
Allow CustomMarking without an __init__() implemented

emmanvg:
Allow no ``__init__()`` for CustomMarking.

### DIFF
--- a/stix2/test/test_custom.py
+++ b/stix2/test/test_custom.py
@@ -94,7 +94,7 @@ def test_custom_property_in_bundled_object():
     assert '"x_foo": "bar"' in str(bundle)
 
 
-def test_custom_marking_no_init():
+def test_custom_marking_no_init_1():
     @stix2.CustomMarking('x-new-obj', [
         ('property1', stix2.properties.StringProperty(required=True)),
     ])
@@ -104,6 +104,8 @@ def test_custom_marking_no_init():
     no = NewObj(property1='something')
     assert no.property1 == 'something'
 
+
+def test_custom_marking_no_init_2():
     @stix2.CustomMarking('x-new-obj2', [
         ('property1', stix2.properties.StringProperty(required=True)),
     ])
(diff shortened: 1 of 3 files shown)


emmanvg:
Related to #109

stix2/v20/common.py
```diff
@@ -145,7 +145,14 @@ class _Custom(cls, _STIXBase):
...
                 _STIXBase.__init__(self, **kwargs)
-                cls.__init__(self, **kwargs)
+                try:
+                    cls.__init__(self, **kwargs)
+                except (AttributeError, TypeError) as e:
+                    # Don't accidentally catch errors raised in a custom __init__()
+                    if ("has no attribute '__init__'" in str(e) or
+                            str(e) == "object.__init__() takes no parameters"):
+                        return
+                    raise e
```
clenk: Codecov is failing because this line isn't tested because none of our tests have a `CustomMarking` with a custom `__init__` function that raises an exception like our `NewType` custom object does in `test_custom.py`. Maybe we should change https://github.com/oasis-open/cti-python-stix2/blob/4990b7edb2e2b9d0377c4fe8ee8d47063f1e5313/stix2/test/test_markings.py#L190 and add a test with `pytest.raises`.

stix2/test/test_custom.py
```diff
@@ -94,6 +94,26 @@ def test_custom_property_in_bundled_object():
...
+    @stix2.CustomMarking('x-new-obj', [
+        ('property1', stix2.properties.StringProperty(required=True)),
+    ])
+    class NewObj():
+        pass
+
+    no = NewObj(property1='something')
+    assert no.property1 == 'something'
+
+    @stix2.CustomMarking('x-new-obj2', [
```
gtback: let's split this into two tests.

stix2/v20/common.py
emmanvg: That line is also not covered for CustomObject (potentially any of the other Custom decorators). https://codecov.io/gh/oasis-open/cti-python-stix2/src/ef3ce9f6f0a678e604bda63f9eb5742ed53f8eff/stix2/v20/sdo.py#L359

stix2/test/test_custom.py
clenk: Then we should also split the analogous tests for CustomObject, CustomObservable, and CustomExtension.

stix2/v20/common.py
clenk: Oh, because our tests' CustomObject's `__init__` raises a `ValueError`, which is not caught here.

emmanvg:
I have updated the tests. Though, it was the only way I managed to pull of raising a TypeError from a custom object... I hope that's enough. I also separated some of the analogous tests for CustomObject, CustomObservable and CustomExtension.

PR #111, opened 2017-11-16
Support de-referencing relationships

### LINKED ISSUE #89: Add functions to get related objects

clenk:
Adds `relationships()` and `related_to()` functions to DataStores, and wrappers to them in `Environment`.

Fixes #89.


clenk:
Should we also move `creator_of()` into DataSource?

stix2/sources/__init__.py
```diff
@@ -191,6 +256,87 @@ def query(self, query=None):
...
 
+    def relationships(self, obj, relationship_type=None, source_only=False, target_only=False):
+        """Retrieve Relationships involving the given STIX object.
+
+        Only one of `source_only` and `target_only` may be `True`.
+
+        Args:
+            obj (STIX object OR dict OR str): The STIX object (or its ID) whose
+                relationships will be looked up.
+            relationship_type (str): Only retrieve Relationships of this type.
```
gtback: Maybe add a note that if not provided, all relationships will be returned, regardless of type.

stix2/sources/__init__.py
```diff
@@ -191,6 +256,87 @@ def query(self, query=None):
...
+        """
+        results = []
+        rels = self.relationships(obj, relationship_type, source_only, target_only)
+
+        try:
+            obj_id = obj.get('id', '')
+        except AttributeError:
+            obj_id = obj
+
+        for r in rels:
```
gtback: Could this be simplified by finding the set of all (unique) IDs that are either source_ref or target_ref of any of the relationships, removing the object itself, then looping over that list? I think this will potentially return multiple copies of the same object, if there are multiple relationships with that object. 

Would it make sense to return this as a dictionary mapping IDs to objects? Which would make it easier for users to construct the (source_obj, rel_type, target_obj) tuples if combined with a separate call to `relationships()`? Or maybe I'm overthinking this.

stix2/sources/__init__.py
```diff
@@ -354,6 +500,102 @@ def query(self, query=None, _composite_filters=None):
...
+            obj_id = obj
+
+        if relationship_type:
+            filters.append(Filter('relationship_type', '=', relationship_type))
+
+        if source_only and target_only:
+            raise ValueError("Search either source only or target only, but not both")
+
+        for ds in self.data_sources:
+            if not target_only:
```
gtback: Why not just call `ds.relationships()` for each item in the list?

stix2/sources/__init__.py
```diff
@@ -191,6 +256,87 @@ def query(self, query=None):
...
+
+        Returns:
+            (list): List of Relationship objects involving the given STIX object.
+
+        """
+        results = []
+        filters = [Filter('type', '=', 'relationship')]
+
+        try:
+            obj_id = obj.get('id', '')
```
gtback: If the dictionary doesn't have an `'id'`, do we want to raise an error quicker? Will having the empty string cause problems down the line (either an error, which would be bad, or invalid results, which would also be bad, or just zero results, which would just be inefficient)?

stix2/sources/__init__.py
```diff
@@ -354,6 +500,102 @@ def query(self, query=None, _composite_filters=None):
...
+        Returns:
+            (list): List of STIX objects related to the given STIX object.
+
+        """
+        if not self.has_data_sources():
+            raise AttributeError('CompositeDataSource has no data sources')
+
+        results = []
+        for ds in self.data_sources:
+            rels = ds.relationships(obj, relationship_type, source_only, target_only)
```
gtback: why not just call `ds.related_to` and combine the results in a set?

stix2/sources/__init__.py
```diff
@@ -58,7 +59,10 @@ def get(self, *args, **kwargs):
                 object specified by the "id".
 
         """
-        return self.source.get(*args, **kwargs)
+        try:
+            return self.source.get(*args, **kwargs)
+        except AttributeError:
+            raise AttributeError('%s has no data source to query' % self.__class__.__name__)
```
gtback: Are the original error messages not clear enough? I agree these are a bit more explicit, but the default `'NoneType' object has no attribute 'get'` isn't too bad, I don't think, when pointing to a line that includes `self.source.get(...)`

stix2/test/test_filesystem.py
```diff
@@ -40,6 +45,25 @@ def fs_sink():
     shutil.rmtree(os.path.join(FS_PATH, "campaign"), True)
 
 
+@pytest.fixture
```
gtback: I don't know if you can use `scope=module` on this type of fixture (with yield), but it would potentially save a bunch of duplicate writing to disk, since the data is read-only. Maybe use a DataSource to make this more explicit?

gtback: I'm actually starting to question whether it makes sense to have filesystem tests at all. None of the `relationships()` code is in the concrete data store classes, so it just seems to be duplicating tests.

stix2/sources/__init__.py
clenk: I think the intent with these error messages was to help the user get to the cause of the issue more quickly. For this PR, I just noticed that while we were checking them on Environment, we weren't on DataStore, and I wanted to be consistent.

This also handles the case where a user creates an Environment without specifying a DataStore or DataSource. It will automatically create a CompositeDataSource with no sources, so if the user tries to do `environ.get()` they'd get an error message about CompositeDataSource rather than Environment.

gtback: Makes sense

stix2/sources/__init__.py
clenk: Premature optimization.  :P I was trying to avoid getting the object id, adding a filter for relationship type, etc., multiple times when we only need to do it once. But I guess it's not that likely that the composite data source would have enough data sources for it to matter too much. But that would make the code cleaner.

Same with `related_to` below.

stix2/sources/__init__.py
clenk: I'm changing it to call `ds.related_to()`, but we can't combine the results in a set until we make STIX objects (`_STIXBase`) hashable.

stix2/sources/__init__.py
clenk: I think that's over-engineering for the `related_to` function, but maybe it could be added as another, separate function.

PR #131, opened 2018-02-26
Taxii collection source 404

### LINKED ISSUE #104: Documentation: CompositeDataSource Examples seems wrong
First the variable ta probably should be something different, since it suggests a Threat Actor.
The 'output' seems to be two indicators - neither one of which has the id 'indicator--37a6a5de-a5b9-425a-903a-4ae9cbf1ff3f

get an object that is only in the filesystem

ta = cs.get('intrusion-set--f3bdec95-3d62-42d9-a840-29630f6cdc1a')
print(ta)

get an object that is only in the TAXII collection

ind = cs.get('indicator--37a6a5de-a5b9-425a-903a-4ae9cbf1ff3f')
print(ind)

### LINKED ISSUE #107: Documentation: Output on Memory page is misleading
This is the output from a loop printing the results of a query - but they look "different"

{'name': 'Urban2', 'created': '2017-09-12T13:26:18.023Z', 'labels': ['rootkit'], 'modified': '2017-09-12T13:26:18.023Z', 'type': 'malware', 'id': 'malware--2daa14d6-cbf3-4308-bb8e-226d324a08e4'}

{
"type": "malware",
"id": "malware--2b3dd412-18a5-4e81-8742-4977068eb3eb",
"created": "2017-09-29T19:52:17.028Z",
"modified": "2017-09-29T19:52:17.028Z",
"name": "Alexios",
"labels": [
"rootkit"
]
}

### LINKED ISSUE #108: Change `allow_custom` in DataStores
- [ ] `allow_custom` should be an attribute on data store/source/sink class instances, ~~and you should be able to overwrite the default in an individual method call. (e.g. you have a DataStore set up with `self.allow_custom = False`, but just this once you want to add an object with a custom property so you call `datastore.add(..., allow_custom=True`)~~
- [ ] `allow_custom` should default to `True` in DataSources; assume that only valid content was placed in the datasource. It can default to `False` in DataSinks.

clenk:
Data source changes.

Fixes #104, fixes #107, fixes #108.


stix2/test/test_memory.py
```diff
@@ -202,64 +202,64 @@ def test_memory_store_save_load_file(mem_store):
...
-        ]
-    }""" % (camp_name, camp_alias, camp_id)
-
-    mem_store.add(bund)
-
-    camp_r = mem_store.get(camp_id)
-    assert camp_r["id"] == camp_id
-    assert camp_r["name"] == camp_name
-    assert camp_alias in camp_r["aliases"]
+# Currently deprecated - removed functionality from Memory API
```
gtback: Are we planning to re-enable this? If not, let's just remove it.

Issue #125, opened 2018-02-02
TypeError: Object of type ThreatActor is not JSON serializable


m0jtaba:
Have I missed something fundamental here, trying to push stix content but getting "TypeError: Object of type ThreatActor is not JSON serializable"

```
from stix2 import TAXIICollectionSink, ThreatActor
from taxii2client import Collection
import json
collection = Collection("http://127.0.0.1:5000/trustgroup1/collections/91a7b528-80eb-42ed-a74d-c6fbd5a26116/", user="admin", password="Password0")

#create TAXIICollectionSINK and push STIX content to it
tc_sink = TAXIICollectionSink(collection)

# create new STIX threat-actor
ta = ThreatActor(name="Teddy Bear",
                labels=["nation-state"],
                sophistication="innovator",
                resource_level="government",
                goals=[
                    "compromising environment NGOs",
                    "water-hole attacks geared towards energy sector",
                ])

tc_sink.add(ta)

**Comment by gtback** [comment]
Hi, @m0jtaba . We clearly have a lot to do in terms of specifying exactly what types of things the sink can accept, and where exactly to draw the line of responsibility between the `Collection` in `taxii2client` and the `TAXIICollectionSink` in `stix2`. I thought I might be able to provide a workaround without hacking into the guts of either library, but I can't figure it out. We do have a plan for reworking both libraries to make this much more straightforward, but until we have a chance to work on that I'm not sure if I have a good answer. Sorry :-( 

@mbastian1135 , @clenk am I missing something?

**Comment by mbastian1135** [comment]
So its been a while and correct me if Im wrong, but tc.add() triggers a conversion to a bundle, then gets cast as a python dict. Then the dict is passed to collection.add_objects() which just passes the dict to a requests post. So is this a json load happening in bundle() (i.e. the sink) or something that is happening in the Collection ? Although I dont know where it would be in Collection... And if its the former why is ThreatActor not serializable?

@m0jtaba Stack trace available?

**Comment by m0jtaba** [comment]
@gtback thanks for taking a look. I just want to make sure that I am clear with what I am trying to achieve.  My goal is to have a TAXII Server receive a STIX2 object from a TAXII Client and currently  I'm using memory backend. Also to convert none STIX2 objects into STIX2 objects using the STIX2 API and pushing the STIX2 object into a TAXII Server via the TAXII Client.


@mbastian1135 see below:
```

python3.4 clientSink.py 

Traceback (most recent call last):
[...]

PR #143, opened 2018-03-16
Update TAXIICollectionSink.add() to use serialize()

emmanvg:
Also call `parse()` on `dict` objects. Pass UTF-8 strings to taxii2client.

### DIFF
--- a/stix2/datastore/taxii.py
+++ b/stix2/datastore/taxii.py
@@ -83,8 +83,7 @@ class TAXIICollectionSink(DataSink):
             # adding list of something - recurse on each
             for obj in stix_data:
                 self.add(obj, version=version)
-            else:
-                return  # If the list is empty just return.
+            return
 
         elif isinstance(stix_data, str):
             # adding json encoded string of STIX content
--- a/stix2/datastore/taxii.py
+++ b/stix2/datastore/taxii.py
@@ -83,6 +83,8 @@ class TAXIICollectionSink(DataSink):
             # adding list of something - recurse on each
             for obj in stix_data:
                 self.add(obj, version=version)
+            else:
+                return  # If the list is empty just return.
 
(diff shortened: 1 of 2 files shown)


gtback:
I'm not sure if there's a good way to add tests for this without having a running TAXII server. Can you think of anything?

treyka:
@gtback You could use https://ubertaxii.com (Medallion-based) as part of your testing harness. Let me know if you need creds and/or new collections added for testing.

stix2/datastore/taxii.py
```diff
@@ -83,6 +83,8 @@ def add(self, stix_data, version=None):
             # adding list of something - recurse on each
             for obj in stix_data:
                 self.add(obj, version=version)
+            else:
+                return  # If the list is empty just return.
```
chisholm: The else clause is executed if no `break` is encountered in the loop causing it to terminate early.  It isn't an empty check.  This loop body is a single line with no break, so it looks to me like the else clause could never be skipped.  It will always run.  So there may as well not be an else clause, right?  Wouldn't
```python
    for obj in stix_data:
        self.add(obj, version=version)

    return
```
do the same thing?

gtback: I agree. I think the code is functioning correctly, but either the comment is outdated or it works for a different reason than what was intended.

emmanvg: I think I can change it for what @chisholm recommends. The intention was to prevent the code to reach `self.collection.add_objects(bundle)` in the list case because bundle would be uninitialized.

emmanvg:
I would suggest to leave the test coverage for another PR. That should give some time to think about possible solution. If there are no further comments about my changes. It should be good to go.

emmanvg:
This fix is related to #125

Issue #136, opened 2018-03-06
When you use the query method, the “type” property should be checked


rpiazza:
Since a STIX type name can only use ‘-’s, we should check to make sure there is no '_' in it (an easy to make typo).

Maybe for all uses of a type name...

clenk:
(This was originally brought up in the context of the `query()` method on DataStores.)

mbastian1135:
@rpiazza - Hey Rich, could you elaborate for me? Does this mean checking the 'type' field in a Filter when it is passed to a DataStore ( or when the FIlter is created)? Or checking the 'type' field of every STIX object searched over?  Or a check when creating a STIX CustomObject?

rpiazza:
Yes - just if it is part of a query.

gtback:
Fixed by #147

Issue #119, opened 2017-12-08
Filter parameters not working for TAXIICollectionSource


johnwunder:
Code:
```
from stix2 import TAXIICollectionSource, Filter
from taxii2client import Collection

# establish TAXII2 Collection instance
collection = Collection("https://limo.anomali.com/api/v1/taxii2/feeds/collections/107/", user="guest", password="guest")

# supply the TAXII2 collection to TAXIICollection
tc_source = TAXIICollectionSource(collection)

# build your filter -- kinda seems broken now
f1 = Filter("type","=", "indicator")

#retrieve the STIX objects
results = tc_source.query([f1], allow_custom=True)
```

That returns:
```
HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: https://limo.anomali.com/api/v1/taxii2/feeds/collections/107/objects/?match%5Bfilters%5D=match%5Btype%5D
```

The error seems to be in generating the filter parameter...it's somehow blowing up and it's formatted incorrectly.

chisholm:
The way stix2 is invoking taxii2-client is wrong.  Looks like stix2 is creating HTTP query parameters itself, and passing them with a "filters" keyword arg; but it should be using a separate keyword arg per parameter, e.g. `collection.get_objects(type="indicator")`. See [here](https://github.com/oasis-open/cti-python-stix2/blob/c0669d7a5f71e73bc040b1114a9649c811ea7ce4/stix2/sources/taxii.py#L212).  taxii2-client does its own conversion [here](https://github.com/oasis-open/cti-taxii-client/blob/f6ceae007849642cfc13614318dbf60419b0a93c/taxii2client/__init__.py#L293).

gtback:
Thanks, @johnwunder. Hopefully we'll have a chance to fix this soon. I've got a lot of ideas for cleaning up the interface between python-stix2 and taxii2-client.

mbastian1135:
Just a note, if the stix2.TAXIICollection interface is altered, we need to alert Anthony as he is writing a blog post using the examples in the docs.

mbastian1135:
Oh wow this is bad, in addition to passing the filters in the wrong format, we arent stripping the TAXII specific filters from the full query. So when we get objects back from TAXII, it looks like we re-apply the TAXII filters to the objects through our filter code, but our filter code throws out filters that are not common properties.  So "version" and "added_after" filters will never return anything.  :( ... My fault, Ill fix

gtback:
[...]

PR #147, opened 2018-03-22
119 taxii filters

mbastian1135:
Covers:

- #119 - Filters that are destined for TAXII server are extracted and sent to taxii2client as keywords. The STIX objects returned from TAXII are then evaluated against the rest of the filters in the query. Filters that are sent to TAXII are not re-evaluated locally against the STIX objects again - doing so causes error as TAXII filter properties are outside the scope of STIX2 common properties and thus would cause STIX objects to be discarded; specifically for filters with properties 'version', 'added_after'. This code was tested against local Medallion server.

- #136 - I put a check for this when Filters are initialized. Its a little incongruous within the context of the other more generic checks. If dont want the check to occur in the Filter initialization, then would have to be when the Filter is applied in _check_filter().

### DIFF
--- a/stix2/datastore/filters.py
+++ b/stix2/datastore/filters.py
@@ -35,7 +35,11 @@ def _check_filter_components(prop, op, value):
 
     if type(value) not in FILTER_VALUE_TYPES:
         # check filter value type is supported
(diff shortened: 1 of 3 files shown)


gtback:
Looks good to me. 👍 to @clenk 's comment about docstrings.

PR #150, opened 2018-03-30
 Create Workbench layer

### LINKED ISSUE #38: Create "Workbench API"
The Workbench API is designed to be a high-level interface to querying, creating, and interacting with STIX data. It builds on the explicit "Environment" concept by exposing a set of module-level functions that operate on an implicit, module-level Environment, without requiring users to be aware of the underlying Environment. (This encapsulation allows applications that need to support multiple Environments to do so, but does not require individual users to manage their own Environment).

clenk:
Fixes #38.

### DIFF
--- a/stix2/datastore/__init__.py
+++ b/stix2/datastore/__init__.py
@@ -99,6 +99,25 @@ class DataStoreMixin(object):
         except AttributeError:
             raise AttributeError('%s has no data source to query' % self.__class__.__name__)
 
+    def query_by_type(self, *args, **kwargs):
+        """Retrieve all objects of the given STIX object type.
+
+        Translate query_by_type() call to the appropriate DataSource call.
+
+        Args:
+            obj_type (str): The STIX object type to retrieve.
+            filters (list, optional): A list of additional filters to apply to
+                the query.
+
+        Returns:
+            list: The STIX objects that matched the query.
+
+        """
+        try:
+            return self.source.query_by_type(*args, **kwargs)
+        except AttributeError:
+            raise AttributeError('%s has no data source to query' % self.__class__.__name__)
+
     def creator_of(self, *args, **kwargs):
         """Retrieve the Identity refered to by the object's `created_by_ref`.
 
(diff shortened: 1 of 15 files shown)


stix2/datastore/__init__.py
```diff
@@ -273,9 +294,32 @@ def query(self, query=None):
                 to conduct search on.
 
         Returns:
-            stix_objs (list): a list of STIX objects
+            list: The STIX objects that matched the query.
+
+        """
+
+    def query_by_type(self, obj_type='indicator', filters=None):
```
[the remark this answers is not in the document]
clenk: Not really. I'd be fine removing the default and making it a required argument instead.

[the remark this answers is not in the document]
emmanvg: I think making it required makes more sense since you can't "query by type" without a type.

stix2/datastore/__init__.py
```diff
@@ -542,6 +596,35 @@ def query(self, query=None, _composite_filters=None):
 
         return all_data
 
+    def query_by_type(self, *args, **kwargs):
```
gtback: It seems like there's a lot of duplication between this function and other functions on the CompositeDataSource. I wonder if there's anything we can do about that. Maybe not in this PR, though.

emmanvg: Maybe along the same lines.... How is this any easier than a regular call to `query()`?

stix2/datastore/filters.py
```diff
@@ -44,6 +44,29 @@ def _check_filter_components(prop, op, value):
...
+        List of Filters.
+
+    """
+    if isinstance(filter_arg, list):
+        filters.extend(filter_arg)
+    else:
+        filters.append(filter_arg)
+
+    return filters
+
```
chisholm: Every call to _assemble_filters() will append more filters to the same list, and return this ever-growing list.  The `filter` parameter shouldn't be defaulted to a mutable value here.

gtback: It looks like most, but not all uses of `_assemble_filters()` pass the `filters` argument, but we shouldn't use a default mutable argument in this function. It's not as bad in the other places, since we don't mutate or return the default list argument.

PR #151, opened 2018-04-02
Fix on MemorySource.load_from_file()

emmanvg:
Simple typo in MemorySource


mbastian1135:
Good catch! Thanks

emmanvg:
The line was uncovered initially. Though, it can be addressed when the datastore coverage fixes are done.

PR #152, opened 2018-04-02
Update stix2 setup configuration

emmanvg:
- Remove `taxii2-client` as a requirement to install `stix2`
- Add `taxii2-client` to the Tox configuration instead
- Re-factor the version and description loading

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,4 +54,7 @@ setup(
         'six',
         'stix2-patterns',
     ],
+    extras_require={
+        'taxii': ['taxii2-client']
+    }
 )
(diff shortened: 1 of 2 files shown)


clenk:
Is there any advantage to adding `taxii2-client` to `extras` so you can do something like `pip install stix2[taxii]`?

emmanvg:
I think it makes sense to have it like that. According to [setuptools](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) you can have recommended dependencies that are not required for all uses of the project.

PR #164, opened 2018-04-12
Allow generic custom observables and custom observable extensions

### LINKED ISSUE #157: Allow generic custom objects for Cyber Observables
The recent changes made to the library allows parsing generic custom objects, but `parse_observable()` does not have the feature implemented.
https://github.com/oasis-open/cti-python-stix2/blob/master/stix2/v20/observables.py#L903 
It also looks like we are **not** using `allow_custom` when we call it.
https://github.com/oasis-open/cti-python-stix2/blob/master/stix2/v20/observables.py#L36

clenk:
Fixes #157. Doing it this way was simpler than the proposal in that issue.

### DIFF
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -6,11 +6,11 @@ import datetime as dt
 
 import simplejson as json
 
-from .exceptions import (AtLeastOnePropertyError, DependentPropertiesError,
-                         ExtraPropertiesError, ImmutableError,
-                         InvalidObjRefError, InvalidValueError,
+from .exceptions import (AtLeastOnePropertyError, CustomContentError,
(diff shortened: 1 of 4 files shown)


emmanvg:
Is `self.__alow_custom` is for all cases of allowing custom or just observables?

clenk:
With this PR, `self.__alow_custom` is only used in observables/extensions, but this could possibly be expanded later. Currently it's only checked in _STIXBase when calling `clean()` on a property.

Putting the `self.__allow_custom` in there where I did just prevents the need to pass `allow_custom` into `_check_property` in several classes.

PR #165, opened 2018-04-13
Allow dictionaries in a ListProperty

### LINKED ISSUE #161: Create an Extension with Dict annidate inside List
Hi,
I'm trying to create a CyberObservable Extension for UserAccount which have to contain a DictionaryProperty() inside a ListProperty(). It is possible? Because when I try to create an extension like this one

```
@CustomExtension(UserAccount, 'ssh_keys', {
   keys': ListProperty(DictionaryProperty(), required=True)
 })
 class SSHKeysExtension:
     pass
```

and use it with example = SSHKeysExtension(keys=[{'test123':123, 'test345','aaaa'}])
I obtain a lot of strange errors (the library seems to interpreter the dict as parameters for __init__()

clenk:
Fixes #161.

### DIFF
--- a/stix2/properties.py
+++ b/stix2/properties.py
@@ -129,6 +129,8 @@ class ListProperty(Property):
                 # constructor again
                 result.append(valid)
                 continue
+            elif type(self.contained) is DictionaryProperty:
+                obj_type = dict
             else:
                 obj_type = self.contained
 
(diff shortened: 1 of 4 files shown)


emmanvg:
For future reference. The dictionary key/value pairs are provided alphabetically. Will explore for future implementations the use of `OrderedDict`. https://stackoverflow.com/questions/6921699/can-i-get-json-to-load-into-an-ordereddict

PR #172, opened 2018-04-26
Fix make_constant when value is already a constant

### LINKED ISSUE #171: make_constant() should check if value is already a constant
``` python
>>> from stix2.patterns import StringConstant
>>> from stix2.patterns import make_constant
>>> a = StringConstant('foo')
>>> make_constant(a)
  File "/home/lu/Projects/cti-python-stix2/stix2/patterns.py", line 166, in make_constant
    raise ValueError("Unable to create a constant from %s" % value)
ValueError: Unable to create a constant from 'foo'
```

clenk:
This fixes a bug discovered in the elevator.

Fixes #171.

### DIFF
--- a/stix2/patterns.py
+++ b/stix2/patterns.py
@@ -147,6 +147,9 @@ class ListConstant(_Constant):
 
 
 def make_constant(value):
+    if isinstance(value, _Constant):
+        return value
+
     try:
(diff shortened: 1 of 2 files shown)


gtback:
Tested this locally with the elevator and it fixes the bug there as well 👍 .

Issue #174, opened 2018-05-02
TAXII Data Source should raise an error on 404


johnwunder:
The following code returns no results, with no error:

```python
from stix2 import TAXIICollectionSource
from taxii2client import Collection

collection = TAXIICollectionSource(Collection("https://ubertaxii.com/taxii/api-root/collections/no-collection-exists/"))

len(collection.query()) # == 0
```

I believe in that case it's an unauthorized error, but this also happens with a 404 and probably some other errors.

It seems like there should be an error raised when a request is made that results in some non-success status from the server, vs. just returning zero results.

gtback:
Related to #113

Issue #113, opened 2017-11-20
TAXIICollectionSource should capture HTTP errors


gtback:
Wrap calls to `get_object` and `get_objects`, catch an HTTP 404 or other errors, and return `None` or an empty list as appropriate.

gtback:
Yep, it looks like it got done. Thinking about this more, though, do we really want to return an empty list for *any* HTTP error? I can understand for a 404, but if the authentication is bad (401, 403), or there's a problem on the server (5xx), do we want to alert the user to that somehow?

gtback:
I'm not sure what you mean by "Graceful API warning response", but I agree that it probably wouldn't be my first choice.

On one hand I don't want to introduce a bunch of custom exceptions, but on the other hand I don't think returning HTTPErrors from requests is the best option either. On the third (?) hand, maybe this is something best handled in the taxii-client, rather than just calling `raise_for_status()` there ([1](https://github.com/oasis-open/cti-taxii-client/blob/573ec9ac5a0aa2432d23683afed96b1463d535f6/taxii2client/__init__.py#L808), [2](https://github.com/oasis-open/cti-taxii-client/blob/573ec9ac5a0aa2432d23683afed96b1463d535f6/taxii2client/__init__.py#L825)).

I'm open to suggestions. @emmanvg  @clenk do you have ideas?

chisholm:
In general, I am okay with returning a special value (e.g. None or empty list) which indicates not-found, instead of raising an exception.  Especially if these not-found conditions are routine and normal, as it can make caller code simpler.  But the spec looks like it defines 404 as meaning either not-found, or access-denied, which makes it ambiguous.  The spec keeps saying for 404s: "... does not exist or the client does not have access ...".  And I wouldn't want to confuse an access-denied error with a not-found error.

Perhaps what the spec intended was to say that it is legal for a TAXII server to make it appear to clients *as if* the requested resource does not exist, when the client doesn't have permission to access it.  The client is still intended to interpret 404's as not-found, in which case the status is effectively unambiguous from the client's point of view.

Would an exception which indicates a 404 status occurred, carry any substantively different information, than an empty list or None, which was documented to mean that the requested resource was not found?

[...]

PR #181, opened 2018-05-18
Http error

mbastian1135:
#113 
#174

### DIFF
--- a/stix2/datastore/__init__.py
+++ b/stix2/datastore/__init__.py
@@ -40,8 +40,9 @@ class DataSourceError(Exception):
 
     def __str__(self):
         if self.root_exception:
-            self.message = "{} \"{}\"".format(self.message, self.root_exception)
-        return self.message
+            return "{} \"{}\"".format(self.message, self.root_exception)
+        else:
+            return self.message
 
 
 class DataStoreMixin(object):
@@ -149,7 +150,7 @@ class DataStoreMixin(object):
             obj (STIX object OR dict OR str): The STIX object (or its ID) whose
                 relationships will be looked up.
             relationship_type (str): Only retrieve Relationships of this type.
-            If None, all relationships will be returned, regardless of type.
+                If None, all relationships will be returned, regardless of type.
             source_only (bool): Only retrieve Relationships for which this
                 object is the source_ref. Default: False.
             target_only (bool): Only retrieve Relationships for which this
--- a/stix2/datastore/__init__.py
+++ b/stix2/datastore/__init__.py
@@ -24,15 +24,35 @@ def make_id():
     return str(uuid.uuid4())
 
 
+class DataSourceError(Exception):
+    """General DataSource error instance, used primarily for wrapping
+    lower level errors
+
+    Args:
+        message (str): error message
+        root_exception (Exception): Exception instance of root exception
+            in the case that DataSourceError is wrapping a lower level or
+            other exception
+    """
+    def __init__(self, message, root_exception=None):
+        self.message = message
+        self.root_exception = root_exception
+
+    def __str__(self):
+        if self.root_exception:
+            self.message = "{} \"{}\"".format(self.message, self.root_exception)
+        return self.message
+
+
 class DataStoreMixin(object):
     """Provides mechanisms for storing and retrieving STIX data. The specific
     behavior can be customized by subclasses.
 
     Args:
         source (DataSource): An existing DataSource to use
-             as this DataStore's DataSource component
+            as this DataStore's DataSource component
         sink (DataSink): An existing DataSink to use
-             as this DataStore's DataSink component
+            as this DataStore's DataSink component
 
     Attributes:
         id (str): A unique UUIDv4 to identify this DataStore.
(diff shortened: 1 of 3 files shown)


stix2/datastore/taxii.py
```diff
@@ -145,9 +179,12 @@ def get(self, stix_id, version=None, _composite_filters=None):
             stix_objs = self.collection.get_object(stix_id)["objects"]
             stix_obj = list(apply_common_filters(stix_objs, query))
 
-        except HTTPError:
-            # if resource not found or access is denied from TAXII server, return None
-            stix_obj = []
+        except HTTPError as e:
+            if e.response.status_code == 404:
```
emmanvg: Would it make sense to add a `logging` statement here. I was looking back to the related issues and @johnwunder was specifically confused about 404 returning an empty list. I know raising the exception is unwanted so maybe logging would at least make you aware that an 404 occured... Any thoughts?

stix2/datastore/taxii.py
```diff
@@ -235,9 +272,12 @@ def query(self, query=None, version=None, _composite_filters=None):
             query.remove(taxii_filters)
             all_data = list(apply_common_filters(all_data, query))
 
-        except HTTPError:
+        except HTTPError as e:
             # if resources not found or access is denied from TAXII server, return empty list
-            all_data = []
+            if e.response.status_code == 404:
```
emmanvg: Is there any reason why this `except` handling is different from the one on line 182?

stix2/datastore/taxii.py
mbastian1135: Ill yield to @gtback on the logging. But again, the logic implemented is pulled from the policy outlined here https://github.com/oasis-open/cti-python-stix2/issues/113.

stix2/datastore/taxii.py
chisholm: I think the idea was that the "singular" getters could return None if the object was not found/status 404.  The "plural" getters raise an error on 404 to disambiguate collection-not-found from collection-was-empty/everything-filtered-out (former uses exception, latter uses empty list).  When looking for a single object, there is no "empty" condition to handle, so it's simpler.  That's my take on it anyway.

stix2/datastore/taxii.py
chisholm: Hmmm... didn't he try a "query()" (plural getter, see my other comment below) on a non-existent collection?  I think that would result in an exception now.  Although I'm never averse to having useful diagnostics :)

stix2/datastore/taxii.py
emmanvg: Read the issue again. Thanks for clarifying.

stix2/datastore/__init__.py
```diff
@@ -129,7 +149,7 @@ def relationships(self, *args, **kwargs):
             obj (STIX object OR dict OR str): The STIX object (or its ID) whose
                 relationships will be looked up.
             relationship_type (str): Only retrieve Relationships of this type.
-                If None, all relationships will be returned, regardless of type.
+            If None, all relationships will be returned, regardless of type.
```
gtback: trivial issue, but I think this should remain indented.

stix2/datastore/__init__.py
```diff
@@ -24,15 +24,35 @@ def make_id():
...
+            in the case that DataSourceError is wrapping a lower level or
+            other exception
+    """
+    def __init__(self, message, root_exception=None):
+        self.message = message
+        self.root_exception = root_exception
+
+    def __str__(self):
+        if self.root_exception:
+            self.message = "{} \"{}\"".format(self.message, self.root_exception)
```
gtback: This is dangerous because if you call it multiple times, the message will keep getting longer.

stix2/datastore/__init__.py
```diff
@@ -24,15 +24,35 @@ def make_id():
...
+
+    Args:
+        message (str): error message
+        root_exception (Exception): Exception instance of root exception
+            in the case that DataSourceError is wrapping a lower level or
+            other exception
+    """
+    def __init__(self, message, root_exception=None):
+        self.message = message
+        self.root_exception = root_exception
```
chisholm: Python3 already supports exception chaining via `__cause__` and `raise x from y` syntax.  Any way we could just use that in python3, instead of reinventing this wheel?

stix2/datastore/__init__.py
mbastian1135: Oversight, I prefer the indent as well.

stix2/datastore/__init__.py
clenk: I think we still need to support python 2 for a little while, unfortunately, but we could change this when that is no longer the case.

chisholm: Six has a [raise_from](https://pythonhosted.org/six/#six.raise_from) function... that would give code compatibility with 2 and 3, but unfortunately that would make the chaining only work under python3.  Well, I have less sympathy for python2 these days, so I went ahead and used it in the matcher.  Maybe it will push people to upgrade :)  Sigh.

PR #185, opened 2018-06-01
fix custom STIX objects with nested dictionary error #184

### LINKED ISSUE #184: Custom STIX objects with nested dictionary error
A custom STIX object with a nested dictionary created using the following example:

``` python
#!/usr/bin/env python3.6

from stix2 import CustomObject, properties

@CustomObject('x-example', [
    ('dictionary', properties.DictionaryProperty()),
])
class Example(object):
    def __init__(self, **kwargs):
        pass

example = Example(dictionary={'key':{'key_one':'value', 'key_two':'value'}})

print(example)
```

throws the following error when calling the `__str__()` method:

``` bash
  File "./example.py", line 16, in <module>
    print(example)
  File "/home/lib64/python3.6/site-packages/stix2/base.py", line 199, in __str__
    return self.serialize(pretty=True)
  File "/home/lib64/python3.6/site-packages/stix2/base.py", line 262, in serialize
    return json.dumps(self, cls=STIXJSONEncoder, **kwargs)
  File "/home/lib64/python3.6/site-packages/simplejson/__init__.py", line 399, in dumps
    **kw).encode(obj)
  File "home/lib64/python3.6/site-packages/simplejson/encoder.py", line 298, in encode
    chunks = list(chunks)
  File "/home/lib64/python3.6/site-packages/simplejson/encoder.py", line 717, in _iterencode
    for chunk in _iterencode(o, _current_indent_level):
  File "home/lib64/python3.6/site-packages/simplejson/encoder.py", line 696, in _iterencode
    for chunk in _iterencode_dict(o, _current_indent_level):
  File "home/lib64/python3.6/site-packages/simplejson/encoder.py", line 652, in _iterencode_dict
    for chunk in chunks:
  File "/home/lib64/python3.6/site-packages/simplejson/encoder.py", line 652, in _iterencode_dict
    for chunk in chunks:
  File "/home/lib64/python3.6/site-packages/simplejson/encoder.py", line 602, in _iterencode_dict
    items.sort(key=_item_sort_key)
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
```

the work around is to call:

``` python
print(example.serialize(pretty=False))
```

I'm assuming its something do with how `simplejson` is trying to sort the keys and struggles with the nested dictionary. Is there a reason why the `__str__()` method defaults to `pretty=True`?

LiamSennitt:
Fixes #184

### DIFF
--- a/stix2/utils.py
+++ b/stix2/utils.py
@@ -165,46 +165,88 @@ def _get_dict(data):
             raise ValueError("Cannot convert '%s' to dictionary." % str(data))
 
 
+def _iterate_over_values(dict_values, tuple_to_find):
+    """Loop recursively over dictionary values"""
+    from .base import _STIXBase
+    for pv in dict_values:
+        if isinstance(pv, list):
+            for item in pv:
+                if isinstance(item, _STIXBase):
+                    val = find_property_index(
+                        item,
+                        item.object_properties(),
+                        tuple_to_find
+                    )
+                    if val is not None:
+                        return val
+                elif isinstance(item, dict):
+                    for idx, val in enumerate(sorted(item)):
+                        if (tuple_to_find[0] == val and
+                                item.get(val) == tuple_to_find[1]):
+                            return idx
+        elif isinstance(pv, dict):
+            if pv.get(tuple_to_find[0]) is not None:
+                try:
+                    return int(tuple_to_find[0])
+                except ValueError:
+                    return len(tuple_to_find[0])
+            for item in pv.values():
+                if isinstance(item, _STIXBase):
+                    index = find_property_index(
+                        item,
+                        item.object_properties(),
+                        tuple_to_find
+                    )
+                    if index is not None:
+                        return index
+                elif isinstance(item, dict):
+                    dict_properties = item.keys()
+                    index = find_property_index(
+                        item,
+                        dict_properties,
+                        tuple_to_find
+                    )
+                    if index is not None:
+                        return index
+
+
 def find_property_index(obj, properties, tuple_to_find):
     """Recursively find the property in the object model, return the index
-    according to the _properties OrderedDict. If it's a list look for
-    individual objects. Returns and integer indicating its location
+    according to the ``properties`` OrderedDict when working with `stix2`
+    objects. If it's a list look for individual objects. Returns and integer
+    indicating its location.
+
+    Notes:
+        This method is intended to pretty print `stix2` properties for better
+        visual feedback when working with the library.
+
+    Warnings:
+        This method may not be able to produce the same output if called
+        multiple times and makes a best effort attempt to print the properties
+        according to the technical specification.
+
+    See Also:
+        py:meth:`stix2.base._STIXBase.serialize` for more information.
+
     """
     from .base import _STIXBase
     try:
-        if tuple_to_find[1] in obj._inner.values():
-            return properties.index(tuple_to_find[0])
+        if isinstance(obj, _STIXBase):
+            if tuple_to_find[1] in obj._inner.values():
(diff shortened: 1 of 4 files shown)


stix2/utils.py
```diff
@@ -165,46 +165,88 @@ def _get_dict(data):
...
+        if isinstance(pv, list):
+            for item in pv:
+                if isinstance(item, _STIXBase):
+                    val = find_property_index(
+                        item,
+                        item.object_properties(),
+                        tuple_to_find
+                    )
+                    if val is not None:
+                        return val
```
chisholm: If this is to be a value-returning function, make sure all code paths return a value.  I think it would return None if control dropped off the end of the function, but still.  There are a lot of conditional returns without an "else" clause.

In fact, since None would be returned if there were no explicit "return", code like:
```python
if val is not None:
    return val
# else, don't return anything
```
means
```python
if val is not None:
    return val
else:
    return None
```
which is exactly the same as:
```python
return val
```
The latter is equivalent, simplest, and ensure there's always an explicit "return".
(Apologies if this is still a work in progress and you just hadn't got around to addressing this yet... :))

stix2/utils.py
```diff
@@ -165,44 +165,88 @@ def _get_dict(data):
...
-                        return len(tuple_to_find[0])
-                for item in pv.values():
-                    if isinstance(item, _STIXBase):
-                        val = find_property_index(item,
-                                                  item.object_properties(),
-                                                  tuple_to_find)
-                        if val is not None:
-                            return val
+        if isinstance(obj, _STIXBase):
+            index = _iterate_over_values(obj._inner.values(), tuple_to_find)
```
gtback: Can just `return` here rather than setting a variable and immediately returning it.

gtback:
I know there wasn't a lot of explicit testing of the `find_property_index` method before, but it's possibly a good candidate for more testing.

LiamSennitt:
I know my 'fix' was a super lazy, but looks like your on top of it. Is there anything I can do, Is it worth me adding the nested dictionary example as a test?

emmanvg:
@LiamSennitt have you signed the individual CLA for [OASIS](https://www.oasis-open.org/resources/open-repositories/cla/individual-cla)?

LiamSennitt:
@emmanvg I have now, just submitted it anyway.

gtback:
@LiamSennitt if you can add a test for the nested dictionary case, that would be great! Anything we can do to improve the code coverage is greatly appreciated. 

@emmanvg (or @LiamSennitt if you have a chance), can you make the couple tweaks that @chisholm and I suggested?

stix2/utils.py
emmanvg: @chisholm, some of the if/else clauses are required because (say a list for example) maybe the first item is not the one I want, so the method would return None. Now on further iteration I can find the object I am searching for and the return a valid index.

stix2/utils.py
emmanvg: These ones are reasonable.

emmanvg:
I think that is all I have in terms of changes. Unless there are more suggestions, it is ready to be merged. I cannot remove some of the checks for `None` because we iterate over a list of values.

stix2/utils.py
gtback: I agree, @emmanvg. In the remaining cases, the `else` is not `return None` but rather "continue to the next item in the list", so I think we're good on this front!

emmanvg: Yeah, I guess I don't know what to return in those cases. My best approach would be to just return something (like a 0) and log it before just to not fail-hard, but that depends if that makes sense as a "solution".

PR #190, opened 2018-06-13
Cannot execute sample code.

stmtstk:
I tried to execute the sample code of parse() on python 2.7 and 3.0.
But It occured JSONDecodeError because the "pattern" value of this sample JSON  are described as multi-line. 
To ensure that the sample code works, this value shuld be single-line.


gtback:
This is just a reformatting change, so no CLA is need.

PR #192, opened 2018-06-15
Revamp code related to "pretty" JSON serialization

chisholm:


emmanvg:
I am ok with the approach. There shouldn't be any more comparison error being raised. Though, when a property is not found the object will simply be printed out of order.

PR #195, opened 2018-06-20
Correct "image_weight" to "image_width".

gtback:
Can you review this, @emmanvg and @clenk . Want to get a second set of eyes on this. Without the change to `stix2/v20/observables.py`, the test will fail. Once we merge this in, we'll need to make the same change in `stix2/v21/observables.py` in the `stix2.1` branch.

### DIFF
--- a/stix2/test/test_observed_data.py
+++ b/stix2/test/test_observed_data.py
@@ -765,6 +765,51 @@ def test_file_example_with_RasterImageExt_Object():
     assert f.extensions["raster-image-ext"].exif_tags["XResolution"] == 4928
 
 
+RASTER_IMAGE_EXT = """{
+"type": "observed-data",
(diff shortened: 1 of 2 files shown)


stix2/test/test_observed_data.py
```diff
@@ -765,6 +765,51 @@ def test_file_example_with_RasterImageExt_Object():
...
+      }
+    }
+  }
+}
+}
+"""
+
+
+def test_raster_image_ext_parse():
+    obj = stix2.parse(RASTER_IMAGE_EXT, allow_custom=False)
```
clenk: `allow_custom` defaults to False so it's not strictly necessary here, but doesn't hurt.

gtback: I'll take it out. I was having problems getting the test to fail initially (it ended up being unrelated), and I thought maybe True was the default for extensions, which is why I added it. I don't want to confuse anyone, though.

PR #198, opened 2018-06-26
update dict key limits per 2.1 spec

treyka:
In the course of testing the i18n code in the 2.1 branch, I noticed that while the `language-content` `contents` property allows 2 char dict keys for specifying languages (as shown in all of the examples and per RFC5646), the 2.1 branch still retained the 2.0 3 char restriction on dict key length.

QED, I split properties.py out v20 vs v21 as was done previously to accommodate other 2.1 changes.

Also, somewhat inexplicably the STIX2 dictionary property limits key length to 250 (*not* 256). While I was mucking about, I fixed this on for 2.1.


gtback:
Thanks, @treyka !

PR #199, opened 2018-06-26
Update base.py

emmanvg:
Fix STIXJSONIncludeOptionalDefaultsEncoder call to super().


gtback:
I just noticed that the line being changed isn't covered by any tests. I assume that's why it didn't cause a problem before. Does it make sense to write a test that uses this class?

PR #201, opened 2018-06-27
Ensure all UUIDs are V4

gtback:
Closes #188

Issue #193, opened 2018-06-18
Consider refactoring how Properties are defined


gtback:
See the conversation in #123

PR #204, opened 2018-07-05
Refactor Properties

gtback:
This is just a few minor changes/improvements I found in the process of working on #201, but wanted to treat them separately. If it makes sense, we could hold off on this and address it as part of #193 .Though at that point it would probably be a major, breaking change, and better to incorporate as part of the STIX 2.1 branch.


emmanvg:
Regarding @gtback comment, I think it would be easier for me to pull from master if this PR gets merged...

PR #209, opened 2018-09-07
Fixed a test fixture to call the cti-taxii-client Collection constructor correctly

### LINKED ISSUE #208: Bring TAXII source/sink/store up to date with cti-taxii-client
Backward-incompatible changes have recently been merged to the cti-taxii-client library.  This has broken some parts of python-stix2.  We need to bring python-stix2 back to compatibility.

One issue I encountered was a result of the modifications for TAXII client issue [#39](https://github.com/oasis-open/cti-taxii-client/issues/39).  The Collection class constructor was changed to accept JSON as a whole object, not broken out into kwargs.  Apparently some test fixtures create Collections in the [old way](https://github.com/oasis-open/cti-python-stix2/blob/b5a301ff28b6a73054741d5cd9a90a0036a7f6e5/stix2/test/test_datastore_taxii.py#L69-L83).  So that breaks some unit tests.

chisholm:
It had been recently changed to address issue 39 in that library.  Fixes #208 .  (I actually thought there might be more breakage besides this, but... this seemed to be the only thing.)

### DIFF
--- a/stix2/test/test_datastore_taxii.py
+++ b/stix2/test/test_datastore_taxii.py
@@ -16,8 +16,10 @@ COLLECTION_URL = 'https://example.com/api1/collections/91a7b528-80eb-42ed-a74d-c
 class MockTAXIICollectionEndpoint(Collection):
     """Mock for taxii2_client.TAXIIClient"""
 
-    def __init__(self, url, **kwargs):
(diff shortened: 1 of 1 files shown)


chisholm:
Oh yeah... tests won't pass until there is a new taxii client release.

clenk:
Looks good, thanks @chisholm!

PR #213, opened 2018-10-03
python 3.7 support

rooterkyberian:
I have tried to update one of my projects to Python 3.7, but right now this is only dependency left that is not yet supporting it, so here is a patch to fix that:)

In case you are wondering what was broken on Python 3.7:
```
In [1]: object(a=1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-5acd2be7f9d9> in <module>()
----> 1 object(a=1)

TypeError: object() takes no arguments
```
the message changed from `no parameters` to `no arguments`. Anyway, I don't think it is a good idea to relay on message text anyway so instead I just check does class redefine `__init__` method or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,11 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "3.7"
+matrix:
+  include:
+    - python: 3.7 # https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905
+      dist: xenial
+      sudo: true
 install:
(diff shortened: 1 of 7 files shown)


emmanvg:
@rooterkyberian have you signed the individual CLA for [OASIS](https://www.oasis-open.org/resources/open-repositories/cla/individual-cla)?

rooterkyberian:
@emmanvg I just did

clenk:
Thank you for the PR @rooterkyberian!

PR #214, opened 2018-10-15
Multi version memory store

### LINKED ISSUE #210: Support multiple versions of an object in MemorySource
This would probably involve storing the objects in a dictionary of lists instead of just a list.

chisholm:
Fixes #210 .

Bug fixes include:
- Objects inside of bundles should be added to memory data sinks, not the bundles themselves.  Doc strings indicated this, and other data sinks did this, but memory data sinks didn't.  Now, they do.  There was one unit test which was actually verifying this erroneous behavior.  I changed the test so that it tested whether the object inside the bundle was queryable.
- The types of data stored in memory sinks was not normalized.  If you added a dict, then a dict was stored.  If you added a stix2 object, the object was stored.  I changed it so that dicts are always parsed into stix2 objects.  So the only thing you can get from memory data sources/stores is stix2 objects now.  This affected several tests, since they were doing timestamp comparisons as strings.  Timestamp property values of stix2 objects are datetime objects, so the string comparisons no longer worked.  I fixed the tests to parse those strings into datetime objects before comparing.
- The "filters" property on memory data sources was not always honored.  E.g. if you call get() to get a latest-version object by ID, and didn't pass a value for _composite_filters, then the filters property was ignored.  If you did pass a value for _composite_filters, then the query was delegated to the query() method, which did honor the filters property.  Now, it always honors the property.
- There have been a few bugs due to the fact that the "allow_custom" setting hadn't been propagated to where it was needed, resulting in errors when users work with custom objects.  This was true of memory stores/sinks: allow_custom was a setting passed into the source/sink constructor, but it was not propagated to the _add() function which is responsible for parsing dict's to stix2 objects when they are added.  Now, allow_custom is propagated appropriately.
- Some unit tests called the store/sink's add() method with a second arg value of True.  But the second arg to add() is a version, and True isn't a legal version.  Not sure why that happened, but I removed it.

Other things:
- The _add function no longer has a trailing else clause which throws a TypeError with a customized message.  This broke one test which was doing a string comparison on the exception message, not just checking that an exception was thrown and what type of exception it was.  But we've recently seen that checking for an appropriate exception based on the message can be brittle (it can change over python releases), and in general I don't think that's an appropriate way to test exceptions.  If it's that important to distinguish one exception from another, we ought to create new exception types, not do string compares on the messages.  So I dropped the message checks.
- I changed the docstring on apply_common_filters().  The function was written pretty generally in terms of what parameter types it supported, which was good, but the docs didn't reflect that generality.  The old docstring said the function required lists and sets, but it didn't, and my code doesn't just pass lists and sets.  I think it should be documented more accurately, and we should aim to maximize generality as much as we reasonably can.  It makes the functions more widely applicable and useful.

### DIFF
--- a/stix2/datastore/filters.py
+++ b/stix2/datastore/filters.py
@@ -123,8 +123,11 @@ def apply_common_filters(stix_objs, query):
     Supports only STIX 2.0 common property properties.
 
     Args:
-        stix_objs (list): list of STIX objects to apply the query to
-        query (set): set of filters (combined form complete query)
+        stix_objs (iterable): iterable of STIX objects to apply the query to
+        query (non-iterator iterable): iterable of filters.  Can't be an
+            iterator (e.g. generator iterators won't work), since this is
+            used in an inner loop of a nested loop.  So we require the ability
+            to traverse the filters repeatedly.
 
     Yields:
         STIX objects that successfully evaluate against the query.
--- a/stix2/datastore/memory.py
+++ b/stix2/datastore/memory.py
@@ -12,16 +12,17 @@ Note:
 
 """
 
+import itertools
 import json
 import os
 
 from stix2.base import _STIXBase
 from stix2.core import Bundle, parse
 from stix2.datastore import DataSink, DataSource, DataStoreMixin
-from stix2.datastore.filters import Filter, FilterSet, apply_common_filters
+from stix2.datastore.filters import FilterSet, apply_common_filters
 
 
-def _add(store, stix_data=None, version=None):
+def _add(store, stix_data=None, allow_custom=True, version=None):
     """Add STIX objects to MemoryStore/Sink.
 
     Adds STIX objects to an in-memory dictionary for fast lookup.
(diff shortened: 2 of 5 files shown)


clenk:
Thanks, @chisholm! I'm still going through this, but calling add() with a 2nd parameter of `True` is from before we reworked `allow_custom` and had to pass it into every function call. Thanks for fixing that.

Could you update the module-level docstring of memory.py? We no longer need the note about only storing a single version of a STIX object.

chisholm:
I removed both TODO's.  The other one regarding deduplicate was also no longer relevant.  The internal data structure doesn't support duplicates since it's basically a map of maps: ID->modified->STIX object.  You get one object per (ID, modified) pair.  Adding the same version of the same object as was previously added, will overwrite previous object.  So it isn't possible for any query to return duplicates.  So there should never be a need to deduplicate.

chisholm:
You know, I was thinking about the filesystem store changes and how I'd mentioned that markings needed to be handled specially.  I realized that the same goes for memory store!  I had assumed when making these memory store changes, that all objects have a "modified" property, but markings don't.  So I have a little more work to do here.

PR #222, opened 2018-11-02
Multi version filesystem store

chisholm:
Added support for storing/retrieving multiple versions of the same object ID.  Restructured stix2 test data corpus, and added some new versions of existing objects to it.  Updated some tests to test multi-versioning.

This also includes significant updates to search optimization:
- The old optimization strategy was pretty conservative.  It has been made a bit more sophisticated.  Examples include:
    * if you had filters ("type", "=", "foo") and ("type", "=", "bar"), the old algorithm would restrict searches to the "foo" and "bar" type directories.  But you could also recognize that no object can have a type property with two different values at the same time, and not search at all.  Now, it will effectively skip the search.
    * if you had filters like ("type", "=", "foo") and ("type", "!=", "foo"), the old algorithm used the "=" filter and ignored "!=".  The new algorithm will recognize that this is an impossible filter.
    * ID filters now affect type-based optimization, since type names appear in IDs.  E.g. if you had ("type", "=", "foo") and ("id", "=", "bar-123..."), this is recognized as an impossible filter.
    * The "in" filter operator is now used for optimizations (one wonders why there isn't a "not in"...)
- The old algorithm did a recursive search through a type directory.  The new algorithm doesn't.  I don't know if it was ever intended that the json files would be spread out over an arbitrarily nested subdirectory structure.
- The old algorithm did name extension checks, but not file type checks (e.g. regular file vs directory).  The new algorithm will check types.  This would ensure, for example, that a directory named "foo.json" wouldn't be mistaken for a regular JSON file.  Not that it's likely that a directory would be named like that...


stix2/datastore/filesystem.py
```diff
@@ -3,13 +3,362 @@
...
+
+from stix2.base import _STIXBase
 from stix2.core import Bundle, parse
 from stix2.datastore import DataSink, DataSource, DataStoreMixin
 from stix2.datastore.filters import Filter, FilterSet, apply_common_filters
-from stix2.utils import deduplicate, get_class_hierarchy_names
+from stix2.utils import get_type_from_id, is_marking
+
+
+def _timestamp2filename(timestamp):
```
emmanvg: Are **any** these methods generic enough or have the ability to be used for other datastores? If this is the case, could we move them to `stix2.utils`, `stix2.datastores.__init__`  or a better location.

chisholm: I did decouple things so that the design felt more like a set of integrated-but-independently-usable and testable pieces (and I did independently test them in the unit tests).  But I think most wouldn't have much applicability outside the filesystem store.  No need for a general white/blacklist class, is there?  The ability to scan a directory in the context of a white/blacklist?  Did you have a particular class/function in mind, that you thought would be useful?

chisholm:
I may be misinterpreting the travis output, but was it an actual check that failed, or some kind of processing hook?  Looks like it was trying to trim whitespace, but that failed.  The files it shows below are the ones which failed?  Or the ones which succeeded?  I don't see any actual error message.

Any idea what that hook actually objected to, and modified?  It seemed to only affect the test data corpus files, and all of those files were created via `json.dump()` or the FileSystemSink, where the file content is determined via `str(stix_obj)`.  None of those were hand-written or edited.  Maybe not worth trying to appease the trailing-whitespace hook...

I'm not sure what happened with these failing checks.

stix2/datastore/filesystem.py
emmanvg: Definitely no need (that I know of) for a generic while/blacklist class or any functionality. Just wondering if there was anything that was re-usable from this code then it would make sense to move it into a move convenient place..

stix2/datastore/filesystem.py
```diff
@@ -77,15 +426,23 @@ def stix_dir(self):
...
-            os.makedirs(os.path.dirname(path))
+        if not os.path.exists(obj_dir):
+            os.makedirs(obj_dir)
 
         if self.bundlify:
             stix_obj = Bundle(stix_obj, allow_custom=self.allow_custom)
 
-        with open(path, "w") as f:
+        with open(file_path, "w") as f:
             f.write(str(stix_obj))
```
emmanvg: I know this was unchanged, but str(stix_obj) is not the right way to serialize objects. You could do something like [this](https://github.com/oasis-open/cti-python-stix2/blob/stix2.1/stix2/datastore/filesystem.py#L93-L95)

emmanvg: Responding to you questions. The tool when you run it locally will tell you which files failed the test and fix them for you. So, the files you see `"Fixing stix2/test/..."` are the problematic ones. The objective of the hook is to prevent extra/unnecessary white-space on the code, but since we pass a "--all-files" flag I think it also consider non-code files like .json.

emmanvg: We don't have to fix the str(stix_obj) thing here since I already have fixes on my other PR branch. It can be taken care on the upcoming merges.

clenk:
Thank you, @chisholm!

PR #229, opened 2018-12-10
Add visitor2

rpiazza:
add fixes for pattern processing

### DIFF
--- a/stix2/STIXPatternVisitor.py
+++ b/stix2/STIXPatternVisitor.py
@@ -3,20 +3,20 @@ import inspect
 
 import six
 from stix2patterns.grammars.STIXPatternLexer import STIXPatternLexer
-from stix2patterns.grammars.STIXPatternParser import (
-    STIXPatternParser, TerminalNode,
-)
+from stix2patterns.grammars.STIXPatternParser import (STIXPatternParser,
+                                                      TerminalNode)
 from stix2patterns.grammars.STIXPatternVisitor import STIXPatternVisitor
 from stix2patterns.validator import STIXPatternErrorListener
-from stix2.patterns import _BooleanExpression
 
 from antlr4 import CommonTokenStream, InputStream
+from stix2.patterns import _BooleanExpression
 
 # need to import all classes because we need to access them via globals()
+from .patterns import (BinaryConstant, BooleanConstant, FloatConstant,
+                       FollowedByObservationExpression, HexConstant,
+                       IntegerConstant, RepeatQualifier, StartStopQualifier,
+                       StringConstant, TimestampConstant, WithinQualifier)
 from .patterns import *  # noqa
-from .patterns import (FollowedByObservationExpression, StartStopQualifier, WithinQualifier,
-                       RepeatQualifier, StringConstant, IntegerConstant, FloatConstant, HexConstant,
-                       BinaryConstant, BooleanConstant, TimestampConstant)
 
 
 def collapse_lists(lists):
--- a/stix2/STIXPatternVisitor.py
+++ b/stix2/STIXPatternVisitor.py
@@ -14,6 +14,9 @@ from antlr4 import CommonTokenStream, InputStream
 
 # need to import all classes because we need to access them via globals()
 from .patterns import *  # noqa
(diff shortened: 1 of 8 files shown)


stix2/STIXPatternVisitor.py
```diff
@@ -0,0 +1,353 @@
...
+    result = []
+    for c in lists:
+        if isinstance(c, list):
+            result.extend(c)
+        else:
+            result.append(c)
+    return result
+
+
+def values_only(things):
```
emmanvg: This method could use better variable names and a better function name in order to be more descriptive.

stix2/STIXPatternVisitor.py
```diff
@@ -0,0 +1,353 @@
...
+
+from antlr4 import CommonTokenStream, InputStream
+from stix2.patterns import _BooleanExpression
+
+# need to import all classes because we need to access them via globals()
+from .patterns import (BinaryConstant, BooleanConstant, FloatConstant,
+                       FollowedByObservationExpression, HexConstant,
+                       IntegerConstant, RepeatQualifier, StartStopQualifier,
+                       StringConstant, TimestampConstant, WithinQualifier)
+from .patterns import *  # noqa
```
emmanvg: Do we need both import statements `from .patterns import *`  and `from .patterns import (...)`?

stix2/STIXPatternVisitor.py
```diff
@@ -0,0 +1,353 @@
...
+                children[0].operands.append(children[2])
+                return children[0]
+            else:
+                return self.instantiate("AndBooleanExpression", [children[0], children[2]])
+
+    # Visit a parse tree produced by STIXPatternParser#propTestEqual.
+    def visitPropTestEqual(self, ctx):
+        children = self.visitChildren(ctx)
+        operator = children[1].symbol.type
+        negated = negated = operator != STIXPatternParser.EQ
```
emmanvg: Seems like `negated` has been duplicated

rpiazza: good catch

stix2/STIXPatternVisitor.py
rpiazza: changed to:

```
def remove_terminal_nodes(parse_tree_nodes):
    values = []
    for x in parse_tree_nodes:
        if not isinstance(x, TerminalNode):
            values.append(x)
    return values
```

stix2/STIXPatternVisitor.py
```diff
@@ -0,0 +1,353 @@
...
+        return self.instantiate("EqualityComparisonExpression", children[0], children[3 if len(children) > 3 else 2],
+                                negated)
+
+    # Visit a parse tree produced by STIXPatternParser#propTestOrder.
+    def visitPropTestOrder(self, ctx):
```
clenk: Looks like (from Travis) that this method isn't being tested. Do we need some tests for it?

stix2/STIXPatternVisitor.py
```diff
@@ -0,0 +1,353 @@
...
+        if isinstance(children[1], StringConstant):
+            # special case for hashes
+            return children[1].value
+        else:
+            return self.instantiate("BasicObjectPathComponent", children[1].getText(), True)
+
+    # Visit a parse tree produced by STIXPatternParser#setLiteral.
+    def visitSetLiteral(self, ctx):
+        children = self.visitChildren(ctx)
+        return self.instantiate("ListConstant", values_only(children))
```
clenk: Can we add a test for this function so `values_only`/`remove_terminal_nodes` gets tested?

stix2/STIXPatternVisitor.py
rpiazza: I added minimal testing for the visitor (just 2) in test_pattern_expressions.py.  I can add more, but I wouldn't hold up the PR

stix2/STIXPatternVisitor.py
rpiazza: Done

stix2/STIXPatternVisitor.py
rpiazza: Done

PR #235, opened 2018-12-18
Additional parameter to avoid failing on parsing

chrisr3d:
As discussed previously on Slack, we wanted to avoid our users to get errors with UUIDs while trying to load their STIX files.

As a result, we added an additional parameter to some of the different functions in order to ensure a certain interoperability between products, considering that some of them could generate time-based UUIDs.


emmanvg:
Hi @chrisr3d, I went back and read through the Slack conversation. After pondering about the subject, I agree with @clenk statements in regards to the OASIS libraries being focused on adhering to spec as written. Therefore, if this is something that gets approved and changed in the spec, I would be ok with making the appropriate modifications to the library.

That said, in the same conversation other members have expressed solutions (e.g., @jordan2175) that would work for your case. If the content and code is something that you control you can easily incorporate UUIDs that are not v4 via custom properties. Once the property name is fixed, others can just check if that field is present/perform the required regex validation and use it accordingly.

emmanvg:
@chrisr3d, I will close this PR due to the statements above. If the situation changes in the future or you understand this needs to be revised feel free to re-open. Thanks!

PR #239, opened 2019-01-07
Remove dictionary/extension property non-empty req

### LINKED ISSUE #238: Creating LanguageContent object has error
If you pass a language content object an empty contents dictionary, you get an exception saying it must be non-empty.  I do not see anywhere in the 2.1 specification that requires this.

```
from stix2.v21 import LanguageContent, Campaign
camp = Campaign(name='foo')
LanguageContent(object_ref=camp.id, object_modified=camp.modified, contents={})
```
produces:
```
  File "/Users/jmg/work/cti-i18n/p/lib/python2.7/site-packages/stix2/base.py", line 161, in __init__
  File "/Users/jmg/work/cti-i18n/p/lib/python2.7/site-packages/stix2/base.py", line 94, in _check_property
    raise InvalidValueError(self.__class__, prop_name, reason=str(exc))
stix2.exceptions.InvalidValueError: Invalid value for LanguageContent 'contents': The dictionary property must contain a non-empty dictionary
```

clenk:
Only bundle.objects and observed-data.objects have a requirement to include at least one item.

Fixes #238.

### DIFF
--- a/stix2/test/v20/test_observed_data.py
+++ b/stix2/test/v20/test_observed_data.py
@@ -1140,6 +1140,16 @@ def test_process_example_windows_process_ext_empty():
     assert excinfo.value.properties == sorted(properties_of_extension)
 
 
+def test_process_example_extensions_empty():
+    proc = stix2.v20.Process(
+        pid=314,
+        name="foobar.exe",
+        extensions={},
+    )
+
+    assert '{}' in str(proc)
+
+
 def test_process_example_with_WindowsProcessExt_Object():
     p = stix2.v20.Process(extensions={
         "windows-process-ext": stix2.v20.WindowsProcessExt(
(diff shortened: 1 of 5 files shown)


stix2/test/v20/test_observed_data.py
```diff
@@ -1140,15 +1140,6 @@ def test_process_example_windows_process_ext_empty():
     assert excinfo.value.properties == sorted(properties_of_extension)
 
 
-def test_process_example_extensions_empty():
```
emmanvg: This test could be re-written to verify the serialization if an empty dictionary is set purposely.

stix2/test/v21/test_observed_data.py
```diff
@@ -1114,15 +1114,6 @@ def test_process_example_windows_process_ext_empty():
     assert excinfo.value.properties == sorted(properties_of_extension)
 
 
-def test_process_example_extensions_empty():
```
emmanvg: This test could be re-written to verify the serialization if an empty dictionary is set purposely.

PR #240, opened 2019-01-09
Fix issue 232

khdesai:


emmanvg:
@khdesai, could we add a test for the FileSystemStore to assert this exception is raised under the correct circumstances?

PR #241, opened 2019-01-10
More optimized way to sort stix2 object properties when calling str()

timngo1:
I was using the stix library for a rare case of creating observed-data objects that could potentially have hundreds or even thousands of observables in the objects section and I noticed that it was taking a very long time to write to a file.

real	12m14.415s
user	12m8.568s
sys	0m0.624s

I took a look on how str() was being handled and my understanding is that when simplejson does its sort, for each object property it does a dfs starting from the root object being printed. This is because the order to print the objects is found by calling obj.object_properties() but the sort function only has the property and so you must search for it every time starting from self.

A better way I found was to just call obj.object_properties on the encoder class itself so that simplejson does not have to continuously search for the property. This made my run significantly faster.

real	0m10.669s
user	0m10.324s
sys	0m0.233s

When running pytest I did see that my way of sorting did cause some more tests to fail so it still needs some tweaking. If you guys like this idea perhaps you can help figure out all the edge cases :)


clenk:
Hi @timngo1, glad you have code that is working for you. However, I'm not convinced that this will work in all versions of Python. For this and the reasons @chisholm mentioned, I'm closing this for now, but I'm open to looking at it again if you have further comments.

PR #258, opened 2019-04-17
Fixes #248

### LINKED ISSUE #248: Ignore empty values for optional fields
Latest version of the library will not allow me to create instances of STIX2 classes with `None` or `[]` values for optional fields.

Example:
```python
In [4]: stix2.Malware(name='test malware', labels=['foo', 'bar'], external_references=[])
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/.p/lib/python3.7/site-packages/stix2/base.py in _check_property(self, prop_name, prop, kwargs)
     89             try:
---> 90                 kwargs[prop_name] = prop.clean(kwargs[prop_name])
     91             except ValueError as exc:

~/.p/lib/python3.7/site-packages/stix2/properties.py in clean(self, value)
    164         if len(result) < 1:
--> 165             raise ValueError("must not be empty.")
    166

ValueError: must not be empty.
```

This forces me to write code like this:
```python
external_references = get_external_references()
if external_references:
    obj = stix2.Malware(
        name='test malware',
        labels=['foo', 'bar'],
        external_references=external_references)
else:
    obj = stix2.Malware(
        name='test malware',
        labels=['foo', 'bar'])
```
which is inconvenient.

Proposed solution: the library should accept empty fields for optional fields but should not serialize these in the output JSON, treating these fields as unset

khdesai:


clenk:
Looks good; can we add a test for it? For example, in the tests for one of the SDOs add a test that creates an object with an empty list for one of the properties, and verify that that property doesn't show up in the result.

clenk:
The object marking test with `[]` actually tests the check that no list properties can be an empty list. With this PR we just drop empty lists so that check won't fire. I think this is fine. If the user passes in an empty list for an optional property, it will just get dropped. If the property is required, they'll get an error saying "no values for required properties" which I think also makes sense.

PR #271, opened 2019-06-25
Update observed-data SDO class for stix2.1

chisholm:
Update observed-data SDO class, adding the new stix2.1 property "object_refs".  Added a couple tests for it.


stix2/v21/sdo.py
```diff
@@ -379,6 +380,10 @@ def _check_object_constraints(self):
             msg = "{0.id} 'last_observed' must be greater than or equal to 'first_observed'"
```
clenk: This constraint has been removed in WD04 because both properties are required.

PR #274, opened 2019-06-28
Avoid issues using parse on bundles

chrisr3d:
As the type of a custom object is dict, there is an issue with the 'parse' function when it is called on a bundle containing custom objects.

Any comment if you see another way to handle this issue is welcome

### DIFF
--- a/stix2/v20/bundle.py
+++ b/stix2/v20/bundle.py
@@ -38,7 +38,7 @@ class Bundle(_STIXBase):
 
     def get_obj(self, obj_uuid):
         if "objects" in self._inner:
-            found_objs = [elem for elem in self.objects if elem.id == obj_uuid]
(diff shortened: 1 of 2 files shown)


clenk:
Hi @chrisr3d, the intent was that custom objects would be defined and registered like the examples [here in the documentation](https://stix2.readthedocs.io/en/latest/guide/custom.html#Custom-STIX-Object-Types). This gives the added benefit of validating instances of your custom class.

Your fix here is small and won't break anything, but there are other places where you'll run into problems if your custom objects are dicts.

emmanvg:
I think this is a good fix. At least for basic functionality of the `get_obj()` method, otherwise the code would throw exceptions in unexpected scenarios when dealing with custom object. We have a similar approach on the datastores implementation because of the custom object handling potential.

PR #283, opened 2019-07-19
Stix2.1 wd05

chisholm:
Major changes include:
- Updates to all SDO classes for STIX 2.1 WD05
- Improved the exception class hierarchy
- Improved how property "cleaning" errors are handled
- Improved how the workbench monkeypatching works


stix2/__init__.py
```diff
@@ -54,7 +54,7 @@
     WithinQualifier,
 )
 from .utils import new_version, revoke
-from .v20 import *  # This import will always be the latest STIX 2.X version
+from .v21 import *  # This import will always be the latest STIX 2.X version
```
emmanvg: This is purposely meant to be `.v20` for now until we make a major release of the library (once STIX 2.1 becomes a CS). The idea is that we are providing the objects to aid the debugging and finding problems on the spec, but the 2.1 is not stable yet. Text from the README follows:

STIX 2.X Technical Specification Support
This version of python-stix2 brings initial support to STIX 2.1 currently at the CSD level. The intention is to help debug components of the library and also check for problems that should be fixed in the specification.

The stix2 Python library is built to support multiple versions of the STIX Technical Specification. With every major release of stix2 the import stix2 statement will automatically load the SDO/SROs equivalent to the most recent supported 2.X Committee Specification. Please see the library documentation for more details.

```suggestion
from .v20 import *  # This import will always be the latest STIX 2.X version
```

rpiazza:
Looks good to me...

stix2/__init__.py
chisholm: I thought the point of the stix2.1 branch was to prepare for that major release.  So everything on that branch should be as if we are preparing for that release.  That's how we test it and make sure it will work.  From that standpoint, the README is wrong and needs to be updated at some point (though that's probably low priority for now).  If people want a version where 2.0 is the default, they should use the master branch.

I had to make this change because the workbench was pulling in all the wrong versions of the stix classes.  Well, wrong on the 2.1 branch because I thought it should default to 2.1 classes, and the v21 workbench test suite definitely needed 2.1 classes.  It seemed clear to me that the intent was to default to 2.1, so it was the v20 workbench test suite that needed to be removed.  If not, we need to reinstate the v20 workbench test suite, and remove the v21 test suite.  But that seems like it defeats the purpose of having a stix2.1 branch.

So... I don't understand why you want to do this.

emmanvg: The plan is that when we merge this, we actually release something like `v1.2.0` which matches the latest (to be approved) STIX 2.1 CSD 02. When STIX 2.1 is approved as a CS we would then change the import statement to load 2.1 objects automatically `v2.0.0` along with any other changes to align our library to the spec.

I am ok with reinstating the v20 tests even if that means we don't test the workbench against v21 for now. Though, I don't know how much or what other areas will be affected by this change.

chisholm: > when we merge this

This means merging stix2.1 to master?  Ok, hadn't realized we were going to make another official minor version release, with the stix2.1 branch merged to master, before STIX 2.1 was finalized.  I thought this branch was to be the next major version of the library, and the first official release with any STIX 2.1 support at all.

> Though, I don't know how much or what other areas will be affected by this change.

Switching test suites wouldn't change anything core to the codebase (as long as I can make it work with the same trick I used with the v21 test suite).  Changing the import probably wouldn't cause widespread breakage either; I didn't run into any issues with the old import until I started looking into the workbench test suite.  I'll give it a shot.

clenk:
This looks great! Can we add some deprecation warnings for Observed Data and for its `objects` property? I think `warnings.warn()` is the preferred way to do this.

Can we add a test for `threat-actor.last_seen` < `first_seen`? Looks like it's the only thing not covered from the coverage report.

Also, (though I would merge the PR without this one) for the testing that was removed because it looked at extension properties, could we get back at least a little of it by putting 'extensions' in the `InvalidValueError.prop_name` and checking that? I think this would mean adding an `except AtLeastOnePropertyError()` to `_check_property`. The tests in question were the ones changed in `test_observed_data.py` for both stix versions.

chisholm:
> Can we add some deprecation warnings for Observed Data and for its objects property? I think warnings.warn() is the preferred way to do this.

Seems to me, the warnings module is primarily for the interpreter and perhaps built-in libraries.  It interacts with the interpreter's `-W` commandline option.  The provision for "user" warnings is via a special exception type, [UserWarning](https://docs.python.org/2/library/exceptions.html#exceptions.UserWarning).  This gives the impression that ordinary users may use the module, but they shouldn't use the other warning types.  Sounds like your idea was to use DeprecationWarning in particular, but I'm not sure that's how it's intended to be used.

> Also, for the testing that was removed because it looked at extension properties

You may have misread the commit comment.  It said: A small amount of testing was removed, since it was looking at custom *exception* properties...

So this isn't specifically about extensions.  I think those changes were regarding exceptions which had previously been TypeError's, and therefore hadn't been caught and replaced with InvalidValueError.  Most of the property presence exception classes fall into that category (at least one, extra, missing, etc).

clenk:
For warnings, I was reading [this](https://www.lesinskis.com/python_deprecation_tutorial.html). Looks like you can set a filter so they show up without the interpreter option, but I'm not sure yet what best practice is for library authors like us.

Yes I misspoke (and probably misread too) about the removed tests. I was trying to get at stuff like the removed code above this: [https://github.com/oasis-open/cti-python-stix2/pull/283/files#diff-5d546fb7e9a4fb90af64f4ae52b50106R799](https://github.com/oasis-open/cti-python-stix2/pull/283/files#diff-5d546fb7e9a4fb90af64f4ae52b50106R799) I was hoping to be able to check a little more precisely that the correct error was encountered.

chisholm:
> ... I'm not sure yet what best practice is for library authors like us.

Hmmm... I couldn't find anything that described an intention for users (of Python interpreters and core libraries, i.e. normal developers) to *only* use UserWarning.  I found a [Deprecation](https://deprecated.readthedocs.io/en/latest/index.html) library, which includes a [white paper](https://deprecated.readthedocs.io/en/latest/white_paper.html) describing ways other Python libraries deal with deprecation warnings, including well-known libraries like Flask and Django.  All signs suggest they use all the warning categories.  Ok then, guess we can too.  I guess that Deprecation library wouldn't really work for us, since we aren't deprecating a function or class (something we can decorate); we're deprecating a virtual "property" whose definition is buried in a data structure... it's more complicated for us.

As far as best practices, I think our libraries should emit the warnings but never set up warning filters.  Any commandline tools we create should be free to set filters or give commandline options to allow user control over them.

> I was hoping to be able to check a little more precisely that the correct error was encountered.

If you throw out the exception object, you can't test it; that's inherent in the original design.  In Python 3, you can chain the original and the new one via a special attribute ([\_\_cause\_\_](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement)), which is what I chose to do.  So on Python 3, you could get the original exception and run some tests.  But they would be python version specific.

Another idea is to not test cleaning via object creation, as that triggers the whole property checking, exception replacing, mechanism.  If you want to test cleaning, directly call clean() on a property.  That will raise the exception directly to you and then you can check whatever you want.

cyvs:
Hi, is there a plan to allow setting deterministic uuidv5 ids for stix sdo objects? 
It seems this is mentioned in the wd05 draft but there are no changes here to address that. 

See: https://github.com/oasis-open/cti-python-stix2/blob/582ba2be2c539ec2ef1f8bb384f5c1deb5e23831/stix2/properties.py#L19

Or is this something that will be included in future PRs?

clenk:
@cyvs Yes, this is something we are working on and will be in a future PR.

chisholm:
The beginnings are already in the stix2.1 branch [here](https://github.com/oasis-open/cti-python-stix2/blob/b1fa177f075653a8e39d6fa1c91bc29db6199dd2/stix2/properties.py#L26).  uuid5's are allowed, but logic for automatically computing one for an SCO from "ID Contributing Properties" is not implemented yet.

PR #284, opened 2019-07-31
removed uuidv4 validation for id property of stix objects

cyvs:
to allows custom uuidv5s to be set (before v2.1 update to stix2 lib)


cyvs:
closing. meant to do this on my local fork

PR #285, opened 2019-08-19
WD 05 SCO

khdesai:


rpiazza:
Is there a reason why the process observable's ID is specified as:

        ('id', IDProperty(_type)),

instead of

        ('id', IDProperty(_type, spec_version='2.1')),

khdesai:
> Is there a reason why the process observable's ID is specified as:
> 
> ```
>     ('id', IDProperty(_type)),
> ```
> 
> instead of
> 
> ```
>     ('id', IDProperty(_type, spec_version='2.1')),
> ```

No reason, that's just a relic of an old way I used to separate the Process SCO from the other SCOs when it came time to creating a deterministic ID (since Process doesn't use deterministic IDs). I've added the spec_version back in. Thanks for catching that.

stix2/base.py
```diff
@@ -125,6 +128,11 @@ def _check_at_least_one_property(self, list_of_properties=None):
                 list_of_properties.remove('type')
         current_properties = self.properties_populated()
         list_of_properties_populated = set(list_of_properties).intersection(current_properties)
+
+        if list_of_properties_populated == set(['id']) and isinstance(self, _Observable):
+            # Do not count the auto-generated id as a user-specified property
+            list_of_properties_populated = None
+
```
chisholm: This implies that if "id" was included in the property group (i.e. that which is subject to the "at least one" presence co-constraint), and that property is the only one present (in an observable), produce an error.  I don't understand the reasoning.  Why include "id" in the group at all?

For that matter, the same goes for the "extensions" property for all objects.  Why would you include that in the group then?

Seems like this ought to be a basic presence co-constraint check, but there's weird special casing going on that's hard to understand.  More comments/docstring would be nice...

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+        try:
+            properties_to_use = self._id_contributing_properties
+            if properties_to_use:
+                streamlined_object = {}
+                if "hashes" in kwargs and "hashes" in properties_to_use:
+                    possible_hash = self._choose_one_hash(kwargs["hashes"])
+                    if possible_hash:
+                        streamlined_object["hashes"] = possible_hash
+                for key in kwargs.keys():
```
chisholm: No need for .keys()

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+                    if key in properties_to_use and key != "hashes":
+                        if type(kwargs[key]) is dict:
```
chisholm: Avoid 'type(...) is <type>'.  Use isinstance(...).  If you don't watch to catch _STIXBase by mistake, choose your order so that check comes before the dict check.  E.g.

if isinstance(x, _STIXBase):
...
elif isinstance(x, dict):
...
else:
...

That also flattens out the if statement.

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+                                else:
+                                    streamlined_object[key] = kwargs[key]
+                        else:
+                            if isinstance(kwargs[key], _STIXBase):
+                                streamlined_object[key] = self._embed_obj_to_json(kwargs[key])
+                            else:
+                                streamlined_object[key] = kwargs[key]
+
+                if streamlined_object:
+                    data = canonicalize(str(streamlined_object), utf8=False)
```
chisholm: If I understand what's going on here, I don't think this is what you meant.  str(...) produces a string.  All this code is going to do is probably escape some junk in the string to make it JSON-compliant.  I think what you meant to do is run their encoder over the whole python structure (of streamlined_object), to produce canonical JSON.

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+                    return required_prefix + str(uuid.uuid5(namespace, str(data)))
```
chisholm: JSON is text.  So `str(data)`, under python3 is redundant.  `data` should already be a string.  Under python2, this will try to encode data using some default encoding, which is likely to fail if there are high-codepoint characters in it.  Looks like the uuid.uuid5() implementation already encodes the given name to UTF-8, so you don't have to do it.  Just pass the name string directly.

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+            return None
+        except AttributeError:
```
chisholm: That's a big complicated chunk of code in the try-except.  I'd like to be able to look inside and understand what attribute you expect not to be there, but there's way too much junk in there to figure it out.  Can you narrow your try-except's to the minimal amount of coverage required?

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+            return {"SHA-1": hash_dict["SHA-1"]}
+        elif "SHA-256" in hash_dict:
+            return {"SHA-256": hash_dict["SHA-256"]}
+        elif "SHA-512" in hash_dict:
+            return {"SHA-512": hash_dict["SHA-512"]}
+        else:
+            # Cheesy way to pick the first item in the dictionary, since its not indexable
+            for (k, v) in hash_dict.items():
+                break
+            return {k: v}
```
chisholm: This will fail if the dict is empty (k and v will not have been assigned values).  I'd prefer something more bulletproof (even if through some logic elsewhere we don't think that will ever happen ;))

Here's a better way to get an arbitrary key from a dict:
```python
k = next(iter(hash_dict), None)
```
This will result in `None` if the dict is empty (of course, we assume None itself wasn't a key!).  You could also get a key/value tuple if you wanted:
```python
kv = next(iter(hash_dict.items()), None)
```
Just be careful to check `kv` for None before trying to get the key and value from the tuple, in case the dict was empty.

stix2/base.py
```diff
@@ -347,6 +360,61 @@ def _check_property(self, prop_name, prop, kwargs):
...
+
+    def _embed_obj_to_json(self, obj):
```
chisholm: Does _embed_obj_to_json() need to be a method?  Doesn't look it uses anything from 'self'.  Could be `@staticmethod`, or a free function.  I think I'd prefer the latter.  It seems almost general-purpose; kind of like _get_dict() but with the defaulted_optional_properties thing.

stix2/base.py
chisholm: Also, it looks like you made a dict to be encoded as (canonical) JSON.  But the STIX spec says:

> The value of the name portion SHOULD be the list of "ID Contributing Properties" ...

Which probably means a *list* of ID Contributing Property *values*.  Not an object.  Does that seem right?  Or did we decide to go against the "should"? :)

stix2/base.py
rpiazza: I think when I originally prototyped this - data was a bytearray not a string...

chisholm: I hadn't realized that this code had been mostly lifted right out of your prototype.  I went back and found it again, and you just had `canonicalize(streamlined_object)`, without the utf8=False part.  So yeah, it would have returned bytes.

stix2/base.py
clenk: The way it's written in the spec seems like it could be interpreted in different ways... I'm okay with either changing it to a list or leaving it a dict until we get more clarification in the spec.

stix2/base.py
```diff
@@ -392,7 +393,12 @@ def _generate_id(self, kwargs):
 
             if streamlined_obj_vals:
                 data = canonicalize(streamlined_obj_vals, utf8=False)
-                return required_prefix + six.text_type(uuid.uuid5(namespace, data))
+                # print (str(type(data)))
+                try:
+                    return required_prefix + six.text_type(uuid.uuid5(namespace, data))
+                except UnicodeDecodeError:
+                    return required_prefix + six.text_type(uuid.uuid5(namespace, six.binary_type(data)))
```
clenk: We need to add a comment as to why we do this (Python 2 compatibility)

PR #287, opened 2019-09-13
Fixing precision not holding on a marking-definition during parsing and serialization

zrush-mitre:


clenk:
Thanks @zrush-mitre!

PR #289, opened 2019-09-16
Semantic Equivalence

emmanvg:
Provides a basis of work for the Semantic Equivalence process.

- Currently lacks support to check for Indicator `pattern`


stix2/environment.py
```diff
@@ -186,3 +192,306 @@ def creator_of(self, obj):
...
+
+        matching_score = 0.0
+        sum_weights = 0.0
+        type1, type2 = obj1["type"], obj2["type"]
+
+        if type1 != type2:
+            raise ValueError('The objects to compare must be of the same type!')
+
+        if obj1.get("spec_version", "2.0") != obj2.get("spec_version", "2.0"):
+            raise ValueError('The objects to compare must be of the same spec version!')
```
clenk: There might be cases where it makes sense to compare across spec versions. Could we only raise this error if there's a property on one object but not on the other that is used by the algorithm from the whitepaper? I think the only properties where this would be the case are the *_types properties. Down the road we might also look into tracking the properties that got renamed, so a 2.0 indicator with a label that matches a 2.1 indicator's indicator_type could be found to be equivalent.

stix2/test/v21/test_environment.py
```diff
@@ -372,3 +377,367 @@ def test_related_to_by_target(ds):
...
+    coa2 = stix2.v21.CourseOfAction(id=COURSE_OF_ACTION_ID, **COURSE_OF_ACTION_KWARGS)
+    ints2 = stix2.v21.IntrusionSet(id=INTRUSION_SET_ID, **INTRUSION_SET_KWARGS)
+    obs2 = stix2.v21.ObservedData(id=OBSERVED_DATA_ID, **OBSERVED_DATA_KWARGS)
+    rep2 = stix2.v21.Report(id=REPORT_ID, **REPORT_KWARGS)
+
+    obj_list = [(coa1, coa2), (ints1, ints2), (obs1, obs2), (rep1, rep2)]
+
+    for obj1, obj2 in obj_list:
+        env = stix2.Environment().semantically_equivalent(obj1, obj2)
+        assert round(env) == 0
```
clenk: Could we be more explicit that we don't support these types and throw an error instead?

emmanvg: An error in particular (perhaps sub-classing STIXError)?

stix2/environment.py
emmanvg: The Note does mention anything about cross-spec comparison or how we could figure out weights for these types of checks. What I could do is have the ability to enable or disable this check and if someone wants to write that kind of logic they could do so without having to modify the main logic.

clenk: If I remember right the Note doesn't mention spec version at all. That approach of making it optional sounds good to me.

stix2/test/v21/test_environment.py
clenk: Probably a new one, subclassing STIXError, called something like `EquivalenceUnsupportedError` or `SemanticEquivalenceTypUnsupportedError`?

stix2/environment.py
emmanvg: Right. I meant to say "The Note does **not** mention..."

stix2/environment.py
```diff
@@ -186,3 +192,306 @@ def creator_of(self, obj):
...
+
+
+def _partial_string_based(str1, str2):
+    from pyjarowinkler import distance
+    return distance.get_jaro_distance(str1, str2)
+
+
+def _custom_pattern_based(pattern1, pattern2):
+    logger.warning("Checking for Indicator pattern equivalence is currently not implemented!")
+    return 0  # TODO: Needs to be implemented
```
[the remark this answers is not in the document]
emmanvg: Sounds reasonable.

stix2/environment.py
```diff
@@ -186,3 +192,306 @@ def creator_of(self, obj):
...
+        Args:
+            obj1: A stix2 object instance
+            obj2: A stix2 object instance
+            weight_dict: A dictionary that can be used to override settings
+                in the semantic equivalence process
+
+        Returns:
+            float: A number between 0.0 and 100.0 as a measurement of equivalence.
+
+        Warnings:
```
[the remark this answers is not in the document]
emmanvg: Done

stix2/test/v21/test_environment.py
emmanvg: Opted for the second option.

emmanvg:
Added both suggestions.

PR #297, opened 2019-10-07
Fix issue 296

khdesai:
Change `invalid_types=""` to `invalid_types=[""]`


khdesai:
This pull request also addresses issues #298 and #299, so please let me know here if those issues have been resolved.

PR #304, opened 2019-11-19
Change semantic equivalence logging

khdesai:
Generalize semantic equivalence-contributing properties checking functionality within environment.py and add prop_scores dict so all scoring results can be in a single python object. 

Current test suite includes tests that do and do not specify a comparison method, so did not add tests at this time. But willing to do so if needed


clenk:
Thanks, @khdesai! These changes have been merged in.

PR #306, opened 2019-11-22
Various fixes

### LINKED ISSUE #305: Update to STIX 2.1WD06
Necessary changes I see:
* [ ] `indicator.pattern_version`: if stix patterning lang used, default is spec_version of the object
* [ ] Empty dictionaries are now prohibited just like empty lists
* [ ] `socket-ext.options` now has some requirements around its keys/values
* [ ] Add a test to make sure we throw an error when trying to create an x509-certificate with only common properties and no x509-specific properties.
* [ ] Add TLSH to `properties.HASHES_REGEX`

zrush-mitre:
Closes #305


stix2/v21/observables.py
```diff
@@ -986,6 +999,26 @@ class X509Certificate(_Observable):
...
+                self.get('signature_algorithm') is None and
+                self.get('issuer') is None and
+                self.get('validity_not_before') is None and
+                self.get('validity_not_after') is None and
+                self.get('subject') is None and
+                self.get('subject_public_key_algorithm') is None and
+                self.get('subject_public_key_modulus') is None and
+                self.get('subject_public_key_exponent') is None and
+                self.get('x509_v3_extensions') is None
+        ):
```
clenk: I think we could use `self._check_at_least_one_property()` here, similar to https://github.com/oasis-open/cti-python-stix2/blob/d4c01157352552e2ec05bcf7b0f79ef96e34022d/stix2/v20/common.py#L49

PR #320, opened 2020-01-08
fix issue 318

### LINKED ISSUE #318: version arg is not passed down with the call to parse in STIXObjectProperty.clean
If the version argument is passed in on the original call to parse, it should be passed down also in any recursive calls.  This was not the case here: https://github.com/oasis-open/cti-python-stix2/blob/e3c2a3a57b83da8ce914ba7ea738d2e2813d551e/stix2/properties.py#L685.

rpiazza:
Fixes #318.

### DIFF
--- a/stix2/properties.py
+++ b/stix2/properties.py
@@ -682,6 +682,6 @@ class STIXObjectProperty(Property):
                 "containing objects of a different spec version.",
             )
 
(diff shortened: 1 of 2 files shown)


clenk:
This has already been fixed by #337.

PR #330, opened 2020-01-28
Ensure new_version() actually creates a new version with a correct "modified" property value.

### LINKED ISSUE #329: new_version() doesn't ensure the new "modified" property is greater than the old one
https://github.com/oasis-open/cti-python-stix2/blob/88426de4241533915ec7f0286254d28d0b03137d/stix2/utils.py#L287

When creating a new version and the user hasn't specifically assigned a new `modified` property value, the current time is used as the new value.  This isn't guaranteed to be any different from the old one.  A unit test failed because the original and new version objects were created so quickly that their timestamps were the same.

chisholm:
Fixes #329 .

Now, if the current timestamp is not later than the old modified property value, the new modified value is set to the old value plus one millisecond.

### DIFF
--- a/stix2/utils.py
+++ b/stix2/utils.py
@@ -285,7 +285,7 @@ def new_version(data, **kwargs):
     cls = type(data)
     if 'modified' not in kwargs:
         old_modified = parse_into_datetime(
-            data["modified"], precision="millisecond"
+            data["modified"], precision="millisecond",
         )
         new_modified = get_timestamp()
         # Ensure the new is newer than the old!
--- a/stix2/utils.py
+++ b/stix2/utils.py
@@ -284,11 +284,14 @@ def new_version(data, **kwargs):
 
     cls = type(data)
     if 'modified' not in kwargs:
-        old_modified = parse_into_datetime(data["modified"])
+        old_modified = parse_into_datetime(
+            data["modified"], precision="millisecond"
+        )
         new_modified = get_timestamp()
         # Ensure the new is newer than the old!
-        if new_modified <= old_modified:
-            new_modified = old_modified + dt.timedelta(microseconds=1000)
+        one_ms = dt.timedelta(milliseconds=1)
+        if new_modified - old_modified < one_ms:
+            new_modified = old_modified + one_ms
         kwargs['modified'] = new_modified
     elif 'modified' in data:
         old_modified_property = parse_into_datetime(data.get('modified'), precision='millisecond')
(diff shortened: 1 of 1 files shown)


stix2/utils.py
```diff
@@ -284,7 +284,12 @@ def new_version(data, **kwargs):
 
     cls = type(data)
     if 'modified' not in kwargs:
-        kwargs['modified'] = get_timestamp()
+        old_modified = parse_into_datetime(data["modified"])
+        new_modified = get_timestamp()
+        # Ensure the new is newer than the old!
+        if new_modified <= old_modified:
+            new_modified = old_modified + dt.timedelta(microseconds=1000)
```
clenk: Is there a reason to use `microseconds=1000` instead of `milliseconds=1`?

chisholm: `timedelta` doesn't have a milliseconds field, but I guess it does support it as a kwarg.  I hadn't realized that.

One thing I did think of though, is that if `new_modified` is later but less than a millisecond later, you may still wind up with a duplicate `modified` field due to loss of precision when it is truncated to milliseconds.  What is really needed is to ensure that `new_modified` and `old_modified` represent different milliseconds.  This is guaranteed if they're >= 1 millisecond different, but not necessarily if they are < 1 millisecond different (but it's still possible).  So it probably actually needs further modification.

chisholm:
I pushed an update to hopefully handle the case when the current timestamp is too close to the old `modified` timestamp, but they are not equal.  It may look funny, but you can have negative timedeltas, and the comparisons work such that the code works, as far as I can tell.  You can't really write a unit test for this, the way the function is currently designed, because the test would have to guarantee that a new version is made within a particular short time period.  So I tried testing it in other ways (pasting code into a separate script where I could play with it more).  We could probably make it more unit-testable if we redesigned it some.

rpiazza:
@chisholm -
Did you see the resolution of https://github.com/oasis-tcs/cti-stix2/issues/219.  It is no longer the case that it must be exactly 3 digits.

chisholm:
As long as this library truncates to 3 decimal digits, we will need to ensure that timestamps are produced which will not accidentally become equal when extra precision is lost.  We may need a separate issue to relax the millisecond precision requirement throughout the codebase.  Part of that will probably be a revision to the design of the `new_version()` function.  But I think this fix should probably be merged so we can ensure correct behavior in the meantime.

clenk:
Superseded by #374.

PR #331, opened 2020-01-28
Remove workaround for "values" being both a Mapping method name and STIX property name

### LINKED ISSUE #324: STIX objects with an optional "values" property can cause crash
There is a workaround which attempts to allow the `values` attribute on STIX object instances, whose name collides with the standard python Mapping method, to work either as a property access or an invocation of the Mapping method.  It was used with the [WindowsRegistryKey SCO](https://github.com/oasis-open/cti-python-stix2/blob/0af1f442c0cb6e30132174b5b657a6a35ef78a08/stix2/v21/observables.py#L941).  Unfortunately, that property is optional and accessed without checking whether it exists, which means when it doesn't exist, invocation as the Mapping method raises a KeyError.

chisholm:
Fixes #324 .

The workaround didn't work (caused crashes under some circumstances).  Now, attributes whose names conflict with Mapping methods will have the Mapping interpretation.  Same-named STIX object properties will not be accessible as attributes.

### DIFF
--- a/docs/guide/creating.ipynb
+++ b/docs/guide/creating.ipynb
@@ -334,6 +334,20 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "-----\n",
+    "### Warning\n",
+    "Note that because these objects are designed to be usable as `Mapping`s, the attributes of that interface retain their mapping meanings and can't be used to access STIX object properties.  This includes the attributes:\n",
+    "- keys\n",
+    "- values\n",
+    "- items\n",
+    "- get\n",
+    "-----"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "Attempting to modify any attributes will raise an error:"
    ]
   },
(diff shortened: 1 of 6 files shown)


clenk:
Looks good. Can we document that any properties with conflicting names can't be accessed this way? I think the place to do so would be in https://github.com/oasis-open/cti-python-stix2/blob/master/docs/guide/creating.ipynb, around in/out cells `[9]` where it talks about accessing the properties like standard Python attributes.

chisholm:
I added a cell with a prominent warning message.  But notice that the notebook metadata had references to python 3 changed to python 2.  Dunno if that's a problem.  That happened because I used python 2 to make the change.  My python 3 stix2 virtualenv was created with python 3.8, which jupyter currently has problems with.  So I could either dump the whole thing and recreate with python 3.7, or just use the python 2-based virtualenv I already had.  It was much easier to do the latter.  But I didn't know it would have that side-effect.  Does that cause any problems?

clenk:
No, there's no issue with the python version change. But the warning gets rendered as an awkward table. Can we match the format of the warning here? https://github.com/oasis-open/cti-python-stix2/blob/b9927fd4a53f77641e59b27737d8497072bc5b91/docs/guide/workbench.ipynb#L772

chisholm:
The table formatting problem seems to have been the fault of [pandoc](https://pandoc.org/), which is what sphinx uses as its format converter.  It has its own [markdown extensions](https://pandoc.org/MANUAL.html#pandocs-markdown), which unfortunately result in ambiguity: it allows 3+ dashes to indicate a horizontal rule, same as standard markdown, but reuses the same syntax for its tables (several dashes above and below).  Jupyter rendered it just fine, so it must use a different system for converting markdown to HTML.  That is unfortunate; now we have to keep two systems happy.

Anyway, I changed it to the other format, and changed the content to not pick on the Mapping methods, since the problem is more general than that.  I didn't give an exhaustive list, only said that accessing certain attributes which collide with method names will return a bound method.

PR #344, opened 2020-02-18
Fix the pattern AST creation function

### LINKED ISSUE #326: Incorrect Pattern Error Propagation
When using `stix2/pattern_visitor.py/create_pattern_object()` with a pattern that does not include matching single quotes (e.g. "[ file: name = 'weirdname]" ), the stix2 library returns an unhelpful and unclear error. This is in comparison to the more useful error message provided by the pattern validator if the pattern was supplied to it directly. Thus, there should be a way for the error(s) generated by the pattern validator to be propagated back through stix2. See below stacktrace for more details.

(Based off of https://github.com/oasis-open/cti-python-stix2/issues/303)

stacktrace

chisholm:
Fixes #326 

It seems like [stix2.pattern_visitor.create_pattern_object()](https://github.com/oasis-open/cti-python-stix2/blob/8aca39a0b037647f055ad11c85e0f289835f9f3d/stix2/pattern_visitor.py#L329) was written based on a copy-paste of a function from the pattern validator.  But it wasn't changed enough.  I fixed it up:

- The docstring still mentioned validation.  That's not what the function does.  It was totally wrong.  I gave it a better one.
- That `pattern[:2]` junk was from the validator checking whether the pattern starts with brackets.  It doesn't belong in the AST builder.  I removed it.
- An error listener was used which produced messages with "FAIL:".  That makes sense in the context of validation, but not here.  There is no success/fail test happening.  I switched to a different error listener which produces a more appropriate message.
- I did an ANTLR parser setup similar to the pattern matcher, with BailErrorStrategy and the appropriate error handling, so that parse errors result in exceptions with sensible error messages.

If we want to make visitor use more convenient, maybe we should give the Pattern class(es) from the pattern validator a `visit()` method analogous to the [walk()](https://github.com/oasis-open/cti-pattern-validator/blob/969b9abf60f3e7874d612a43ca2f94b7d4593702/stix2patterns/v21/pattern.py#L36) method?

An example of error output now (with Python 3):
```
  File "c:\programming\python\stix\pattern-validator\stix2patterns\v20\grammars\STIXPatternParser.py", line 2090, in primitiveLiteral
    raise NoViableAltException(self)
antlr4.error.Errors.NoViableAltException: None

The above exception was the direct cause of the following exception:

  File "testx_pattern_parse.py", line 3, in <module>
    errs = create_pattern_object(pattern="[ file: name = 'weirdname]")
  File "C:\Programming\python\stix\cti-python-stix2\stix2\pattern_visitor.py", line 363, in create_pattern_object
    six.raise_from(
  File "<string>", line 3, in raise_from
stix2patterns.exceptions.ParseException: 1:15: no viable alternative at input '''
```

### DIFF
--- a/stix2/pattern_visitor.py
+++ b/stix2/pattern_visitor.py
@@ -10,6 +10,7 @@ from stix2patterns.grammars.STIXPatternParser import (
     STIXPatternParser, TerminalNode,
 )
 from stix2patterns.grammars.STIXPatternVisitor import STIXPatternVisitor
+from stix2patterns.v20.pattern import Pattern
 
 from .patterns import *
 from .patterns import _BooleanExpression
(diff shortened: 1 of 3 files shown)


clenk:
Looks great! I think we should have a test case to make sure the right exception class is being thrown.

I wouldn't be opposed to also adding a visit() method.

chisholm:
I added the visit() method and changed `create_pattern_object()` to take advantage of it.  Of course, travis tests won't pass until a new version of the pattern validator is released which has the change.  (And we too are suffering from the ANTLR 4.8 issue.)

PR #345, opened 2020-02-18
Fixes #323

### LINKED ISSUE #323: Explain id_contributing_properties in custom content documentation
(for 2.1)

khdesai:
Add documentation for 2.1 SCO deterministic IDs and their `_id_contributing_properties`


clenk:
This issue was meant to have the documentation explain how users would specify what properties are ID-contributing when registering their custom SCOs. So ideally it would be added to the `Custom Cyber Observable Types` section and show an example of registering a custom SCO and defining a `_id_contributing_properties` list in the class. We can leave the existing

clenk:
Depends on #354.

clenk:
Merged in with 67548a8, thanks @khdesai!

PR #353, opened 2020-02-25
Update exception message in an indicator unit test

### LINKED ISSUE #352: Indicator unit test broke due to an exception message change
https://github.com/oasis-open/cti-pattern-validator/pull/71 changed an exception message which one of the unit tests of this library was [checking](https://github.com/oasis-open/cti-python-stix2/blob/c2b71672f5504b975b113e3671e3279ceb503a65/stix2/test/v21/test_indicator.py#L274).  The unit test needs an update.

chisholm:
Fixes #352 

I still question whether we should be checking exception messages at all.

### DIFF
--- a/stix2/test/v21/test_indicator.py
+++ b/stix2/test/v21/test_indicator.py
@@ -271,7 +271,7 @@ def test_indicator_stix20_invalid_pattern():
         )
 
     assert excinfo.value.cls == stix2.v21.Indicator
-    assert "FAIL: The same qualifier is used more than once" in str(excinfo.value)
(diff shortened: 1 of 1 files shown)


chisholm:
This is fixing a compatibility problem with the master branch pattern validator.  Recent changes on that branch haven't been officially released, so the unit tests Travis runs are using an older version of the library.  So unit tests without this PR won't break until the next pattern validator release is made, and the tests with this PR won't pass until then either.  I guess you could consider this a proactive PR, fixing a bug which *will* appear when the next pattern validator release is made.

Btw, it could be made to work in either case by not checking the message at all.  An exception was raised before and it's raised now, so that hasn't changed.

PR #355, opened 2020-02-26
Changed several *_types properties to be optional due to STIX spec change

### LINKED ISSUE #335: Because of "unknown", several properties have been changed to optional
See oasis-tcs/cti-stix2#211

chisholm:
Fixes #335 

This changes several properties which were formerly required, to be optional.  Unit tests were updated too, where it seemed to be appropriate.  Test changes were a judgement call, and not all object types had the same types of tests, so not all object type unit test suites seemed to need changes.


clenk (approved):
LGTM!

PR #358, opened 2020-03-02
Add swid property to the software SCO

### LINKED ISSUE #349: Changes made to Software SCO, and its usage.
See https://github.com/oasis-tcs/cti-stix2/issues/220

chisholm:
Fixes #349

### DIFF
--- a/stix2/test/v21/test_observed_data.py
+++ b/stix2/test/v21/test_observed_data.py
@@ -1300,7 +1300,7 @@ def test_software_example():
     s = stix2.v21.Software(
         name="Word",
         cpe="cpe:2.3:a:microsoft:word:2000:*:*:*:*:*:*:*",
-        swid="e2489b43-a436-47fd-9ea5-165ff0316114",
+        swid="com.acme.rms-ce-v4-1-5-0",
         version="2002",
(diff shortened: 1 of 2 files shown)


clenk:
I think the format of swid tags is more like `org.fedoraproject.Fedora-29`. Should we update the test example? I don't know the swid tag for MS Word, though.

rpiazza:
@clenk, @chisholm  - looked at examples in https://nvlpubs.nist.gov/nistpubs/ir/2016/NIST.IR.8060.pdf

Here is a good one: 
tagId="com.acme.rms-ce-v4-1-5-0".

Maybe change the cpe one to correspond to this?

chisholm:
I was looking for info about tagIds too when I was developing this.  I don't remember where I saw it, but I remember reading that they aren't required to be any particular thing.  There may be some "should" rules around them though.  So a UUID seemed good enough, and not technically wrong.  But I'll change the test.

rpiazza:
Yes - I agree - its not defined very well.  It took the text for the STIX spec directly from the document I referred to above.

clenk:
Thanks @chisholm!

PR #362, opened 2020-03-05
Add parent_directory_ref as an ID contributing property for file SCOs

### LINKED ISSUE #361: Need parent_directory_ref as an ID contributing property of file SCOs
From: https://github.com/oasis-tcs/cti-stix2/issues/200

chisholm:
Fixes #361 

I didn't add a unit test for this, just added the property name to the contributing property list.  There were no unit tests for specific SCOs, just the general deterministic ID machinery, so perhaps it isn't necessary.


clenk:
Thanks @chisholm!

PR #370, opened 2020-03-11
New STIX 2.1 SCO extension name requirement: must end with "-ext"

### LINKED ISSUE #364: Enforce requirement that SCO extension names must end with "-ext"
Related: oasis-tcs/cti-stix2#216

chisholm:
Fixes #364 

I made more changes than you may think was strictly necessary, but I thought some aspects of the custom SCO extension code needed some improvement.

- Most requirements on SCO extensions were checked in `_custom_extension_builder`, which is the implementation for the `CustomExtension` decorators (both 2.0 and 2.1).  The former called `_register_observable_extension` to enable the extension.  So you could bypass most STIX2 requirements by bypassing the decorator and builder function and registering an extension directly.  I decided that was undesirable, so I moved all of the requirement checking into the register function.
- Some of the aforementioned checking was for some reason embedded inside the dynamically created extension subclass (`_CustomExtension`), which didn't make any sense and cluttered up the class.  So a side effect of the above is that the class is a lot cleaner and clearer now.
- The custom extension decorator's `properties` had previously been documented as requiring a list of tuples.  That seemed unnecessarily restrictive, so I changed the check to a more sensible one (I thought), and changed the exception message for a violation.  Really, all you ever needed was something you can turn into a dict via `OrderedDict(...)`, which includes any mapping, as well as iterables of length-2-iterables (e.g. tuples of tuples, lists of lists, etc).  One might argue that allowing unordered mappings means property order might be unpredictable.  But if that's important, people don't have to use those types (and we never do that).  I think the added flexibility is fine.
- Changing the exception message broke several unit tests which were checking message contents.  I removed the message checks.  The important thing is that the exception is being thrown.
- Since the SCO extension name syntax rules are now different between STIX 2.0 and 2.1, I had to ensure that both rulesets were checked as appropriate.  The existing code had a regex; I added a special second regex for 2.1+ SCO's which hopefully enforces the correct requirements (it's basically the same as the old one but with "-ext" at the end).  The extension registration function now has to check the STIX version and choose the right regex to use.
- The `version` default changed in the registration function.  I needed a version so I could choose the right SCO extension name regex.  `None` doesn't fit the bill, but the `stix2.DEFAULT_VERSION` does, and is immutable.  The code already behaved that way, but didn't use the obvious implementation, so I simplified it and made my life easier.
- Many 2.1 unit tests used SCO extension names which became invalid, so I had to make up valid ones.  I tried to make them similar to the old ones.

### DIFF
--- a/stix2/core.py
+++ b/stix2/core.py
@@ -7,10 +7,10 @@ import re
 
 import stix2
 
-from .base import _STIXBase
+from .base import _Observable, _STIXBase
 from .exceptions import ParseError
 from .markings import _MarkingsMixin
-from .utils import _get_dict
+from .utils import SCO21_EXT_REGEX, TYPE_REGEX, _get_dict
 
 STIX2_OBJ_MAPS = {}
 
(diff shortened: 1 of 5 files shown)


chisholm:
Hmmm... and now that I go back and look at that, I see there were several customization functions which basically worked the same way (object, observable, marking, ...) and I only changed one of them.  So now it lacks symmetry.  Maybe I should change all the rest too?  :)

clenk:
This is definitely cleaner; I like it! Custom extension names still have a SHOULD start with "x-" though - could we add that back into the v21 tests so they follow best practices?

clenk:
Thanks @chisholm!

PR #374, opened 2020-03-17
Support STIX 2.1 version precision

### LINKED ISSUE #350: Constraint on number of digits in created and modified properties has changed
See https://github.com/oasis-tcs/cti-stix2/issues/219

Has implications for https://github.com/oasis-open/cti-python-stix2/issues/329

chisholm:
Fixes #350 , #329 

The new rules say versioning timestamps need *at least* millisecond precision; the old ones said *exactly* millisecond precision.  So we are able to allow more precision in those timestamps now.  This PR adds support for expressing and enforcing a minimum precision on versioning timestamps.


chisholm:
There is a test failure specific to Python 3.6.  I think it might be a result of:

https://bugs.python.org/issue32404

My unit test uses `utcnow()` whereas the bug report describes `fromtimestamp()`, but I think both methods invoke the same underlying code which contains the bug.  This is not something we can fix.  All we can do is work around it for that particular unit test and version of Python.

Issue #368, opened 2020-03-06
Custom property names must match [a-z0-9_]


clenk:
This applies to custom properties on STIX objects, SCOs, and extensions, including custom objects and extensions. WD07 extended this requirement to custom extensions but it was already on STIX Objects and custom objects though we haven't been enforcing this.

PR #375, opened 2020-03-19
Issue 365

rpiazza:
#365 Type and property names must begin with a letter 
#368 Custom property names must match [a-z0-9_]

### DIFF
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -16,9 +16,11 @@ from .exceptions import (
     MissingPropertiesError, MutuallyExclusivePropertiesError,
 )
 from .markings.utils import validate
-from .utils import NOW, find_property_index, format_datetime, get_timestamp
+from .utils import (
+    NOW, PREFIX_21_REGEX, find_property_index, format_datetime, get_timestamp,
+)
 from .utils import new_version as _new_version
-from .utils import revoke as _revoke, PREFIX_21_REGEX
+from .utils import revoke as _revoke
 
 try:
     from collections.abc import Mapping
@@ -172,8 +174,10 @@ class _STIXBase(Mapping):
             if self.get_class_version() == "v21":
                 for prop_name, prop_value in custom_props.items():
                     if not re.match(PREFIX_21_REGEX, prop_name):
-                        raise InvalidValueError(self.__class__, prop_name,
-                                                reason="Property name '%s' must begin with an alpha character." % prop_name)
+                        raise InvalidValueError(
+                            self.__class__, prop_name,
+                            reason="Property name '%s' must begin with an alpha character." % prop_name,
+                        )
 
         # Remove any keyword arguments whose value is None or [] (i.e. empty list)
         setting_kwargs = {}
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -173,7 +173,7 @@ class _STIXBase(Mapping):
                 for prop_name, prop_value in custom_props.items():
                     if not re.match(PREFIX_21_REGEX, prop_name):
                         raise InvalidValueError(self.__class__, prop_name,
-                                                reason="Property names must begin with an alpha character.")
+                                                reason="Property name '%s' must begin with an alpha character." % prop_name)
 
         # Remove any keyword arguments whose value is None or [] (i.e. empty list)
         setting_kwargs = {}
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -169,10 +169,16 @@ class _STIXBase(Mapping):
             extra_kwargs = list(set(kwargs) - set(self._properties))
             if extra_kwargs:
                 raise ExtraPropertiesError(cls, extra_kwargs)
-        if custom_props:
+        else:
+            # because allow_custom is true, any extra kwargs are custom
+            extra_kwargs = list(set(kwargs) - set(self._properties))
+
+        if custom_props or extra_kwargs:
             self._allow_custom = True
             if self.get_class_version() == "v21":
-                for prop_name, prop_value in custom_props.items():
+                all_custom_prop_names = extra_kwargs
+                all_custom_prop_names.extend(list(custom_props.keys()))
+                for prop_name in all_custom_prop_names:
                     if not re.match(PREFIX_21_REGEX, prop_name):
                         raise InvalidValueError(
                             self.__class__, prop_name,
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -2,6 +2,7 @@
 
 import copy
 import datetime as dt
+import re
 import uuid
 
 import simplejson as json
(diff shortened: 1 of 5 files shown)


stix2/base.py
```diff
@@ -76,6 +79,11 @@ def get_required_properties(properties):
 class _STIXBase(Mapping):
     """Base class for STIX object types"""
 
+    def get_class_version(self):
+        module_name = self.__class__.__module__
+        module_parts = module_name.split(".")
+        return module_parts[1]
+
```
chisholm: Yuuuuck.  An ugly dependency between STIX version and what module a class is in.  I had a similar need in `new_version`, where I needed to know which STIX version the object was so I could use the correct versioning precision rules.  But in that case, the object was fully formed and I could use the `spec_version` property.  Here, it seems you're not so lucky.

Could we just make all 2.1 objects extend a special 2.1 class and use isinstance()?  And the same for 2.0 objects?

stix2/base.py
```diff
@@ -163,6 +171,13 @@ def __init__(self, allow_custom=False, **kwargs):
                 raise ExtraPropertiesError(cls, extra_kwargs)
         if custom_props:
             self._allow_custom = True
+            if self.get_class_version() == "v21":
+                for prop_name, prop_value in custom_props.items():
+                    if not re.match(PREFIX_21_REGEX, prop_name):
+                        raise InvalidValueError(
+                            self.__class__, prop_name,
+                            reason="Property name '%s' must begin with an alpha character." % prop_name,
+                        )
```
chisholm: I think this only checks custom properties when the object was created with the `custom_properties` parameter.  You can also give the property directly and include `allow_custom=True`, and then the property name wouldn't be checked.  So I think this check needs to be done in a different way.  For example:

```python
f = stix2.v21.File(name="foo.exe", _a_b_=11, allow_custom=True)
```

produces no error.  But:

```python
f = stix2.v21.File(name="foo.exe", custom_properties={"_a_b_": 11})
```

produces an error.

stix2/core.py
```diff
@@ -286,11 +319,12 @@ def _register_observable_extension(
                 ext_type,
             )
     else:  # 2.1+
-        if not re.match(SCO21_EXT_REGEX, ext_type):
+        if not re.match(TYPE_21_REGEX, ext_type):
```
chisholm: This is for custom SCO extensions: shouldn't the extension name be matched to `EXT_21_REGEX`?

stix2/utils.py
```diff
@@ -25,8 +26,10 @@
 # STIX object properties that cannot be modified
 STIX_UNMOD_PROPERTIES = ['created', 'created_by_ref', 'id', 'type']
 
-TYPE_REGEX = r'^\-?[a-z0-9]+(-[a-z0-9]+)*\-?$'
-SCO21_EXT_REGEX = r'^\-?[a-z0-9]+(-[a-z0-9]+)*\-ext$'
+TYPE_REGEX = re.compile(r'^\-?[a-z0-9]+(-[a-z0-9]+)*\-?$')
+TYPE_21_REGEX = re.compile(r'^\-?([a-z][a-z0-9]*)+(-[a-z0-9]+)*\-?$')
+EXT_21_REGEX = re.compile(r'^\-?([a-z][a-z0-9]*)+(-[a-z0-9]+)*\-ext$')
+PREFIX_21_REGEX = re.compile(r'^[a-z].*')
```
chisholm: If the 2.1 rule is that all type names must start with a letter, then `TYPE_21_REGEX` shouldn't start with `^\-?`.  That allows matching strings to start with a hyphen.  Maybe `EXT_21_REGEX` too, if extension names count as "types".

stix2/base.py
rpiazza: @clenk , @chisholm - I like this idea.  What I did was the best I could come up with after discussing it with Chris.  Let's make it independent of this PR, and I can use it once it is implemented and tested

stix2/base.py
rpiazza: I was more concerned about testing with a leading numeric character - which would be impossible to do using the first call above.  I wonder if this should be an illegal property name anyway.

stix2/utils.py
rpiazza: I didn't add the `\-?`.  It was part of the re before I made any changes.  I guess we should determine if it should be there.  This is related to a comment above...

stix2/core.py
rpiazza: Good catch - fixed

rpiazza:
Doesn't section 3.1 supersede 11.1?

emmanvg:
Actually 11.2 is the section that supersedes it (my bad). I think we can drop my second recommendation for Custom Objects type names. Found the requirement:

- The type property in a Custom Object **MUST** be in ASCII and **MUST** only contain the characters a–z (lowercase ASCII), 0–9, and hyphen (-).
- The type property **MUST NOT** contain a hyphen (-) character immediately following another hyphen (-) character.

stix2/base.py
clenk: Agreed; we should use a 2.1 class with `isinstance()`, and fix it in a subsequent issue/PR.

stix2/base.py
clenk: Yes, we need to enforce this with `allow_custom=True` too.

stix2/utils.py
clenk: `\-?` shoudn't be there for 2.1.

clenk:
> * I think the type value check can be performed via the `properties.py` module and TypeProperty `class`. This was it is consistent with all other properties allowing: first a spec_version= optional parameter, and a check inside only if the requirement needs to be satisfied for a particular spec_version. This would require all `v21` objects to add the new argument. This helps to remove checking in the module name to determine its version.

I think @chisholm's solution is simpler because we won't have to add a new parameter all over the place. Unless there's a hidden gotcha I'm not thinking of right now.

stix2/utils.py
```diff
@@ -25,8 +26,10 @@
 # STIX object properties that cannot be modified
 STIX_UNMOD_PROPERTIES = ['created', 'created_by_ref', 'id', 'type']
 
-TYPE_REGEX = r'^\-?[a-z0-9]+(-[a-z0-9]+)*\-?$'
-SCO21_EXT_REGEX = r'^\-?[a-z0-9]+(-[a-z0-9]+)*\-ext$'
+TYPE_REGEX = re.compile(r'^\-?[a-z0-9]+(-[a-z0-9]+)*\-?$')
+TYPE_21_REGEX = re.compile(r'([a-z][a-z0-9]*)+(-[a-z0-9]+)*\-?$')
+EXT_21_REGEX = re.compile(r'([a-z][a-z0-9]*)+(-[a-z0-9]+)*\-ext$')
```
chisholm: We still need the leading `^` on the regexes to ensure they match the whole string.

PR #377, opened 2020-03-23
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy.

### DIFF
--- a/stix2/environment.py
+++ b/stix2/environment.py
@@ -364,7 +364,7 @@ def partial_string_based(str1, str2):
 
     """
     from rapidfuzz import fuzz
-    result = fuzz.token_sort_ratio(str1, str2)
+    result = round(fuzz.token_sort_ratio(str1, str2))
     logger.debug("--\t\tpartial_string_based '%s' '%s'\tresult: '%s'", str1, str2, result)
     return result / 100.0
 
--- a/stix2/environment.py
+++ b/stix2/environment.py
@@ -363,8 +363,8 @@ def partial_string_based(str1, str2):
         float: Number between 0.0 and 1.0 depending on match criteria.
 
     """
-    from fuzzywuzzy import fuzz
-    result = fuzz.token_sort_ratio(str1, str2, force_ascii=False)
+    from rapidfuzz import fuzz
+    result = fuzz.token_sort_ratio(str1, str2)
     logger.debug("--\t\tpartial_string_based '%s' '%s'\tresult: '%s'", str1, str2, result)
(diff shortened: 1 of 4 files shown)


stix2/environment.py
```diff
@@ -363,8 +363,8 @@ def partial_string_based(str1, str2):
         float: Number between 0.0 and 1.0 depending on match criteria.
 
     """
-    from fuzzywuzzy import fuzz
-    result = fuzz.token_sort_ratio(str1, str2, force_ascii=False)
```
maxbachmann: rapidfuzz never enforces ascii

maxbachmann:
Does pass all tests now for all supported python versions https://github.com/maxbachmann/cti-python-stix2/runs/537510575
@clenk are python2.7 and 3.4 support something that should be kept in future releases?

clenk:
Thank you for bringing this to our attention, @maxbachmann. We'll be dropping 2.7 and 3.4 in the next release, so this should be fine: [latest travis build](https://travis-ci.org/github/oasis-open/cti-python-stix2/builds/667408373).

Are you able to sign the [contributor license agreement](https://www.oasis-open.org/resources/open-repositories/cla/individual-cla) per [CONTRIBUTING.md](https://github.com/oasis-open/cti-python-stix2/blob/master/CONTRIBUTING.md)? This isn't exactly code, but it would still be great if you could.

stix2/environment.py
emmanvg: I don't think rounding the result is neccesary.

maxbachmann: This is done because fuzzywuzzy returns only full numbers between 0 and 100 (rounds the results before returning), which is not done by rapidfuzz (returns a float between 0 and 100).
So when this is not required, the rounding can be  removed.

clenk:
I've confirmed with @chet-ensign that this is minor enough to not require a CLA. Thanks.

maxbachmann:
@clenk as far as I remember I did sign the CLA anyways :)

PR #386, opened 2020-04-09
Use a fast ISO8601 date time parser

2xyo:
# Problem to Solve

Stix parsing is slow.

# Troubleshooting 

Profiling `stix2.parse`:

```py
from stix2 import parse
def test():
    with open('enterprise-attack.json') as f:
        parse(f, allow_custom=True)

    with open('pre-attack.json') as f:
        parse(f, allow_custom=True)

    with open('stix-capec.json') as f:
        parse(f, allow_custom=True)
```
41% of execution time is used by `dateutil.parser` :
image

Python `dateutil.parser.parse()` is slow! 

# Proposed Solution 

Try string parsing into a `datetime` object with `ciso8601`, a fast ISO8601 date time parser for Python written in C. 

As `ciso8601` only supports the most common subset of ISO 8601, if it fails, try again with the slow `dateutil.parser.parse()`. 

# Performance 

```py
from stix2 import parse

def test():
    with open('enterprise-attack.json') as f:
        parse(f, allow_custom=True)

    with open('pre-attack.json') as f:
        parse(f, allow_custom=True)

    with open('stix-capec.json') as f:
        parse(f, allow_custom=True)

if __name__ == '__main__':
    import timeit
    print(timeit.timeit("test()", number=10, setup="from __main__ import test"))

```

Before:
```console
$ python testcase.py 
16.522664724034257
```
After:
```console
$ python testcase.py 
37.800530530977994
```

Profiling after (without timeit):

image

# Additional Information
* https://github.com/closeio/ciso8601

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
     keywords='stix stix2 json cti cyber threat intelligence',
     packages=find_packages(exclude=['*.test', '*.test.*']),
     install_requires=[
+        'ciso8601',
         'enum34 ; python_version<"3.4"',
         'python-dateutil',
         'pytz',
--- a/stix2/utils.py
+++ b/stix2/utils.py
@@ -14,6 +14,7 @@ from dateutil import parser
 import pytz
 import six
 
+from ciso8601 import parse_datetime
 import stix2
 
 from .exceptions import (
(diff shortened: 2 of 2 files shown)


clenk:
Hi @2xyo, thank you for your contribution! I agree parsing is a bit slow. However, I would rather not introduce a dependency that requires compiling C code to install, and ciso8601 doesn't provide pre-compiled wheels. We've looked at some alternative libraries but they seem to have similar C compilation requirements, haven't been maintained for a couple years, or aren't really much faster. For now we will switch to using the default Python datetime.

Let us know if you think this will be unsatisfactory. Thanks again for bringing this up!

2xyo:
Hi @clenk , ciso8601  provides pre-compiled wheels now :) see  https://github.com/closeio/ciso8601/issues/25#issuecomment-1362087093

PR #392, opened 2020-05-13
Fix issue 389

khdesai:


rpiazza:
I had implemented this somewhat differently - Here is what I did:

       if properties_to_use:
            streamlined_object = {}
            if "hashes" in original_json_object and "hashes" in properties_to_use:
                possible_hash = choose_one_hash(original_json_object["hashes"])
                if possible_hash:
                    streamlined_object["hashes"] = possible_hash
            for key in original_json_object.keys():
                if key in properties_to_use and key != "hashes":
                    streamlined_object[key] = original_json_object[key]
            data = canonicalize(streamlined_object)

I think this makes more sense - because I think this is the way others would implement it.

What do you think?

PR #393, opened 2020-05-13
resolve problem with SSDEEP use in hashing-algorithm-ov

### LINKED ISSUE #391: SSDEEP Hashing algorithm case mismatch
Currently if an `ssdeep` hash is added to an object it will change it to its lowercase variant (not aligned with Section 10.6 STIX 2.1 https://docs.oasis-open.org/cti/stix/v2.1/cs01/stix-v2.1-cs01.html#_tumklw3o2gyz) because stix2 will "clean" the field name. This causes the validator to emit a warning on a "unrecognized hash".

Validator:
``` 
[!] Warning: file--xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx: {241} Object 'file--xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx' has a 'hashes' dictionary with a hash of type 'ssdeep', which is not a value in the hash-algorithm-ov vocabulary nor a custom value prepended with 'x_'.
```

https://github.com/oasis-open/cti-python-stix2/blob/df92770d255d66bd614d4975fc4602c7407998fb/stix2/properties.py#L420

emmanvg:
Eliminates problem in `stix2` where an object using an ssdeep hash in 2.1 would use the lowercase variant. closes #391


emmanvg:
@clenk I rebased, and build passed now...

PR #397, opened 2020-05-20
Drop python-dateutil and switch to built-in datetime module

chisholm:
Dateutil proved too slow.

Most of the changes in this PR are in the unit tests, and are due to them trying to use the format flexibility that dateutil had.  I removed tests/parameter values which were specifically testing for that, since with this PR, we are dropping that flexibility.  In one case, values were missing their "Z"s, which seemed more like laziness, so I added them.

On unexpected side-effect of the switch is that it errors now when there is too much precision to the right of the decimal point.  There were a couple unit tests which verified the old behavior which didn't cause an error; I changed the tests to verify the errors (so it's checking for a raised exception now).  I don't know if anyone relied on it working with excessive precision before.

Another unexpected result is that the pattern AST code had apparently relied on the flexibility to understand a STIX 2.1 timestamp literal format (e.g. `t'1111-22-33T44:55:66.77Z'`, with the `t` prefix) without any modification.  That doesn't work anymore, so I had to add a check for that syntax and strip the surrounding junk off so it would work.  That's done in the visitor.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,6 @@ setup(
     packages=find_packages(exclude=['*.test', '*.test.*']),
     install_requires=[
         'enum34 ; python_version<"3.4"',
-        'python-dateutil',
         'pytz',
         'requests',
         'simplejson',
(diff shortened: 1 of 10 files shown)


rpiazza:
@chisholm -

I think this might work so you don't need two patterns:

"%Y-%m-%dT%H:%M:%S(.%f)?Z"

Issue #396, opened 2020-05-19
SCO references using objects vs IDs cause different deterministic ID generation


maybe-sybr:

maybe-sybr:
Here's an example of the nested identifiers using `archive-ext`:

```python
import stix2.v21

ad = stix2.v21.Directory(path="archived/path")
f_obj = stix2.v21.File(name="foo", extensions={
    "archive-ext": {
        "contains_refs": [ad, ],
    }
})
f_ref = stix2.v21.File(name="foo", extensions={
    "archive-ext": {
        "contains_refs": [ad.id, ],
    }
})

print(f_obj.serialize(pretty=True))
print(f_ref.serialize(pretty=True))
```
Outputs:
```
{
    "type": "file",
    "id": "file--7e26ad71-b261-5289-a8e3-3d7126321233",
    "name": "foo",
    "extensions": {
        "archive-ext": {
            "contains_refs": [
                "{\n    \"type\": \"directory\",\n    \"id\": \"directory--c1af83bc-164d-546d-8a73-207f7ad87324\",\n    \"path\": \"archived/path\",\n    \"spec_version\": \"2.1\"\n}"
            ]
        }
    },
    "spec_version": "2.1"
}
{
    "type": "file",
    "id": "file--916b26ba-65fd-5d95-95d7-866d24c9f98e",
    "name": "foo",
    "extensions": {
        "archive-ext": {
            "contains_refs": [
                "directory--c1af83bc-164d-546d-8a73-207f7ad87324"
            ]
        }
    },
    "spec_version": "2.1"
}
```

Interestingly, this also exposes another handling bug where the referenced Directory isn't rendered by its ID! So the deterministic IDs for the Files are different, and the serialised File which referenced the Directory object was incorrectly serialised.

maybe-sybr:
Thanks for all these fixes, @clenk and @chisholm . Much appreciated!

PR #402, opened 2020-06-02
Fix deterministic ids

### LINKED ISSUE #395: `CustomExtension`s for `File` with `TimestampProperty`s can't handle `datetime` objects
STIX 2.1 File SCOs require that their extensions be considered when generating deterministic UUIDs, serialised following JSON canonicalisation rules. Defining a `CustomExtension` for `Files` that include a timestamp property results in exceptions being raised when that property is provided as a `datetime` object rather than a Zulu formatted datestamp string. The following snippet demonstrates the issue:

```py
import datetime
import pprint

import stix2.v21
import stix2.properties

@stix2.v21.CustomExtension(stix2.v21.File, "x-xtime-ext", (
    ("xtime", stix2.properties.TimestampProperty(), ),
))
class XTimeExtension():
    pass

try:
    f = stix2.v21.File(
        name="foo",
        extensions={"x-xtime-ext": {"xtime": datetime.datetime.utcnow()}, },
    )
except Exception as exc:
    print(exc)
else:
    raise Exception("That should have failed...")

f = stix2.v21.File(
    name="foo",
    extensions={"x-xtime-ext": {"xtime": str(datetime.datetime.utcnow())}, },
)
s = f.serialize()
pprint.pprint(s)
stix2.parse(s)

f = stix2.parse("""
{"type": "file",
 "spec_version": "2.1",
 "name": "foo",
 "extensions": {
   "x-xtime-ext": {"xtime": "1970-01-01T00:00:00Z"}
 }
}
""")
print(f)
```

Presumably the fix would involve teaching `_Observable._generate_id()` to use a `STIXJSONEncoder` rather than just calling `stix2.canonicalization.Canonicalize.canonicalization()` which uses the default `JSONEncoder` implementation.

chisholm:
Fixes #395, #396 .

Revamped deterministic ID generation code:
- "Primary" object initialization is now done first, and the resulting property values are what are used to create the deterministic ID instead of whatever kwargs had been passed into the constructor.
- Conversion to a JSON-serializable object (as preparation for canonicalization) should now handle all types of property values.

    The way this conversion should be done isn't clear from the spec.  E.g. it says to use the "list of ID Contributing Properties" and create canonical JSON from that, but (a) not all property values are compatible with JSON, and (b) it doesn't describe what from each property to use (name+value or just value?).  For (a), property values which are incompatible with JSON are first serialized according to the JSON serialization rules for the type.  This produces a string, which is JSON compatible and so can be canonicalized.  This notably applies to timestamps, for which the library uses instances of a datetime subclass for convenience, which will get their spec-prescribed textual format.  For (b), the decision is that both names and values are important, so we create a canonical JSON *object*, rather than taking the word "list" as given in the spec, literally.


rpiazza:
Can you see any issues with using refs as contributing properties?  There are a few objects that use refs.  I know I have to be careful about the order in which I create the ids when creating the deterministic ids by hand, but maybe it would never be an issue using python-stix2.

chisholm:
By hand, I guess you'd have to create the referent object IDs first, then you can populate the ref properties with those IDs and then canonicalize.  It is similar with the stix2 library: you need *some* value to use for the ref property, so you have to create the referent object (or its ID) first.

Circular references are impossible if both SCOs require UUIDv5's.  You'd have to switch one to a v4 to make that work.

PR #404, opened 2020-06-03
Fix star-imports from stix2.v20 and v21

### LINKED ISSUE #399: Star import of `v20` shadows `stix2.base`
The star import from `v20` in `stix2/__init__.py` causes the `stix2.v20.base` module to be aliased at `stix2.base`, which shadows the actual `stix2/base.py` module meaning it can't be accessed from external code (and is probably order of execution dependent from internal imports). This makes it hard to access the `STIXJSONEncoder` class from my code which has been a bit of a pain. For the moment, I just comment out the star import since I'm using explicit imports from `v21` anyway.

chisholm:
Fixes #399 .

stix2/\_\_init\_\_.py currently does a star-import from stix2.v20.  Unfortunately that brings in a "base" symbol which becomes stix2.base, and hides the actual stix2/base.py module.  This PR adds `__all__`s to the v20 and v21 modules which explicitly lists what symbols should be imported, so we don't get "base" by mistake from them.

(Impl note: the text block split() technique makes it easier to keep the \_\_all\_\_ lists up-to-date, because you can just copy-paste the symbol lists from the imports above.)

### DIFF
--- a/stix2/v20/__init__.py
+++ b/stix2/v20/__init__.py
@@ -103,3 +103,33 @@ EXT_MAP = {
         'unix-account-ext': UNIXAccountExt,
     },
 }
+
+
+# Ensure star-imports from this module get the right symbols.  "base" is a
+# known problem, since there are multiple modules with that name and one can
+# accidentally overwrite another.
+__all__ = """
+    Bundle,
+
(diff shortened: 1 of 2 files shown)


clenk:
Thanks @chisholm!

PR #411, opened 2020-06-12
Fix bug with custom objects inside List and Reference properties

### LINKED ISSUE #333: A Custom Object in object_refs of a Report Object.
Hi,

I appended a custom object ref in object_refs of a Report object and wrote into a file.
However stix2 library can not parse the stix2 file.

I checked the source code and found the line below.

` ('object_refs', ListProperty(ReferenceProperty(valid_types=["SCO", "SDO", "SRO"], spec_version='2.1'), required=True)),
`
It seems that an object_refs property allows to contain only  "SCO", "SDO", "SRO", but a custom object does not realized as those.

Is is the right behavior? 
Or is a Custom Object setting wrong?

If you have any idea, please let me know.

Thank you!

khdesai:
Fixes #333.


clenk:
In `ListProperty.clean()` we check if the List contains `ReferenceProperties` but there are other cases where list properties contain items that need `allow_custom`. I think we also need to account for `ExternalReference`, `GranularMarking`, `KillChainPhase`, `EmbeddedObjectProperty`.

chisholm:
It looks to me like this ReferenceProperty.clean() implementation allows any ID type if allow_custom is True.  That would even allow things explicitly disallowed by the spec.  E.g. the spec says relationship/source_ref can't refer to another relationship.  I think this should be enforced even if allow_custom=True.

I think this concern is specific to cases where `invalid_types` is used, i.e. we define legal types via a "blacklist".  Those types should continue to be invalid.  What allow_custom=True ought to mean for ReferenceProperty with an object type blacklist is that we will allow non-spec-defined types to be used, while still maintaining the blacklist.

Does this make sense?  allow_custom is not intended to allow users to be noncompliant with the spec.  It's intended to enable users to disallow customization at those points where the spec otherwise allows it.  Customization must still follow spec rules.

In fact, I don't think allow_custom makes sense at all here in the "whitelist" case.  When the spec says ref must be to an object of type X, that means exactly what it says.  That is not a valid customization point.  When it says ref must not be to an object of type X, there is an implied customization point there, in the sense that it may refer to a custom object.  Nevertheless, it may *never* refer to an object of type X.

clenk:
I agree with @chisholm's comment. Also, I think in `test_parse_obs_with_custom_obj_refs_allow_custom_true` and `test_parse_obs_with_custom_obj_refs_allow_custom_false` the only thing that should be custom is the body_raw_ref. The test with allow_custom true has 2 custom types in `_valid_refs`, and the test with allow_custom false has the custom ref in the email message instead of the embedded `body_multipart`. Want to make sure we're just testing the custom stuff in `body_multipart`.

chisholm:
A followup to this: perhaps it's still not clear what to do about the original issue.  Looks like it was about the Report SDO and allowing custom object types in the object_refs property.  I think allowing any type when a type whitelist is given is definitely wrong (regardless of allow_custom).  What do we do about the Report property then?  The spec describes it as: "Specifies the STIX Objects that are referred to by this Report."  That's not really whitelisting *or* blacklisting anything.  So here's an idea: for properties like that, use an empty blacklist.  That means nothing is disallowed, or in other words everything is allowed.  Since it's a blacklist, it activates customization enforcement like it needs to.  And it does not have unintended consequences (as far as I can see) which can compromise spec compliance when normal non-empty whitelists or blacklists are used.  I think with my implementation (on my branch with the customization enforcement revamp), it will do the right thing.  Perhaps it could be made to work here too.

PR #418, opened 2020-06-24
Drop support for Python versions older than 3.5

clenk:


tox.ini
```diff
@@ -42,8 +42,6 @@ commands =
 
 [travis]
 python =
-  2.7: py27, style
```
emmanvg, line 45: Do we want to run the `style` commands under a different Python env?

clenk: We already are for 3.8.

emmanvg: lol didn't see it.

PR #424, opened 2020-07-08
Semantic Equivalence - partial_string_based()

emmanvg:
Remove round call from `fuzz.token_sort_ratio(str1, str2)` call in partial_string_based function.


clenk:
LGTM.

Issue #429, opened 2020-07-15
Create final bundle (i.e. enterprise-attack.json)


payamsamimi:
Hi guys, 

recently, I am working on a python recipe, which should generate a final json dataset such as enterprise-attack.json for our attack navigator. I thought, it could be an easy task, thank to this library. However, it changed to be a nightmare. 
Using the query function of Filter class, I am able to gather all entries from a domain. However, once I want to gather all of them in a bundle and write to the disc, I have various problems. 

Issue:
1- filtering tactics returns dict. I have read many of comments, solved issues as well as documentation, still I cannot create a bundle with tactics.

2- When I ignore the customObjects, adding other stix objects cause:
Invalid value for Bundle 'objects': This property may only contain a non-empty dictionary or object 

3- When I use Sink to push to the disc, instead of creating an entry in the appropriate folder, a new folder with an unique ID is created, and a json file with a random name in it. 

I would appreciate, if you could help me.

```python

# reading the domain
src = stix2.FileSystemSource(stix_dir="./attack", allow_custom=True, encoding='utf-8')

tactics = src.query([stix2.Filter('type','=','x-mitre-tactic')])
tecs = src.query([stix2.Filter('type','=','attack-pattern')])
rel = src.query([stix2.Filter('type','=','relationship')])
course_of_actions = src.query([stix2.Filter('type','=','course-of-action')])
identity = src.query([stix2.Filter('type','=','identitiy')])
adversaries = src.query([stix2.Filter('type','=','intrusion-set')])
malwares = src.query([stix2.Filter('type','=','malware')])
marking_definition = src.query([stix2.Filter('type','=','marking-definition')])
tools = src.query([stix2.Filter('type','=','tool')])
matrix = src.query([stix2.Filter('type', '=', 'x-mitre-matrix')]);


final_bundle = Bundle(tactics, tecs, rel, course_of_actions,identity, adversaries, malwares, marking_definition, tools, matrix )


print(final_bundle)
```

clenk:
Hi @payamsamimi, sorry for the difficulties. Looking at your comment on #270, I think some of the problem is that the ATT&CK content uses STIX 2.0 but this library now defaults to using STIX 2.1. You can fix this by specifically importing the 2.0 version of the classes.

For example if you register the custom class:

```python
from stix2 import Filter, FileSystemSource, properties
from stix2.v20 import CustomObject, Bundle

[...]

PR #431, opened 2020-07-20
Fix bug when adding custom object to FileSystemSink if the object type hasn't been registered

clenk:
Related: #429.


emmanvg:
General comment for this PR is that serialization logic being removed from `_STIXBase` and being standalone it can follow a similar path to the `versioning.py` or `parsing.py` modules. Perhaps we could move the `serialize`, `STIXJSONEncoder` and the `STIXJSONIncludeOptionalDefaultsEncoder` to their own separate module.

This would also remove the `__all__` override under `base.py`

chisholm:
While you're at it... what about the `find_property_index()` stuff from utils?  It's only used for pretty-serialization.  Maybe it belongs in serialization.py too.

emmanvg:
Agree @chisholm, will make the change

clenk:
These changes look good to me!

PR #439, opened 2020-07-27
fix: Respect name of `@Custom*` decorated defs

maybe-sybr:
This change makes it easier to refer to `@Custom*` decorated classes programmatically.

### DIFF
--- a/stix2/custom.py
+++ b/stix2/custom.py
@@ -35,6 +35,8 @@ def _custom_object_builder(cls, type, properties, version, base_class):
             base_class.__init__(self, **kwargs)
             _cls_init(cls, self, kwargs)
 
+    _CustomObject.__name__ = cls.__name__
+
     _register_object(_CustomObject, version=version)
     return _CustomObject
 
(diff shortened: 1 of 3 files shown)


maybe-sybr:
This is a fairly minor change but if you need a CLA signed, let me know. Thanks!

clenk:
Hi @maybe-sybr. Thank you for the submission, it's a good change. Would you mind fixing the unit tests though: [https://travis-ci.org/github/oasis-open/cti-python-stix2/builds/712041095](https://travis-ci.org/github/oasis-open/cti-python-stix2/builds/712041095). If you're able to sign a CLA that would be awesome: [https://www.oasis-open.org/resources/open-repositories/cla/individual-cla](https://www.oasis-open.org/resources/open-repositories/cla/individual-cla)

Thanks again!

maybe-sybr:
@clenk I've sent in a CLA, lmk if there are any issues. In the meantime, I've also fixed up those two unit tests which were failing and CI went green at https://travis-ci.org/github/oasis-open/cti-python-stix2/builds/712398792 .

I did notice that the class names were only explicitly references for `CustomExtension`s and more as a side-effect of another behavioural test. Is it at all useful to add tests specifically for properties of the derived class objects? Since we're not passing up things like `__doc__` as well, we're still not really wrapping the decorated class properly, but that may also not be a thing we particularly care about.

clenk:
Thanks, @maybe-sybr your CLA's on file. Yes, it would be useful to have tests for these properties specifically. Properly wrapping the decorated class and passing up all of the attributes hasn't been a priority for us but would be an improvement. We'd accept any pull requests to that end, though! :)

PR #444, opened 2020-08-19
Update CompositeDataSource and deduplicate() support for unversioned objects

### LINKED ISSUE #441: Many datastores and helpers assume created and modified fields exist in objects
As an example, `CompositeDataSource`s attempt to access the `modified` key directly as well as making use of the `deduplicate()` helper which attempts to access `modified` and then falls back to `created` if that's missing. The following snippet shows a handful of breakages which present as you squash each one.

```python
import traceback

import pytest
import stix2.datastore
import stix2.datastore.memory
import stix2.v21

m1 = stix2.datastore.memory.MemoryStore()
m2 = stix2.datastore.memory.MemoryStore()
c = stix2.datastore.CompositeDataSource()
c.add_data_sources((m1.source, m2.source))

o = stix2.v21.Software(name="foo")
m1.add(o)
m2.add(o)

with pytest.raises(KeyError) as caught:
    c.get(o.id)
print(caught.getrepr())
```

The table in section 3.2 shows the exceptions, but even that's not reliable since custom properties could be implemented per section 11.1 to implement versioning per section 3.6 (e.g. for SCOs).

This grep line shows all accesses outside the tests, but I suspect only `utils.py`, `versioning.py` and the datastores would need to be changed.
```sh
$ grep -Er "\[[\"'](created|modified|revoked)" stix2/*.py stix2/canonicalization/ stix2/confidence/ stix2/datastore/ stix2/markings/ stix2/v20/ stix2/v21/
...
```

Most of these accesses seem to be related to sorting versions so it seems like it might be reasonable to instead implement a single helper in `versioning.py` which accepts a series of objects, sorts and (optionally?) deduplicates them. I'm happy to work on a patchset for this but wanted to get some input on an approach.

chisholm:
Fixes #441 .

Update CompositeDataSource and deduplicate() to handle unversioned objects, including 2.1 SCOs.

### DIFF
--- a/stix2/datastore/__init__.py
+++ b/stix2/datastore/__init__.py
@@ -481,14 +481,14 @@ class CompositeDataSource(DataSource):
             if data:
                 all_data.append(data)
 
-        # remove duplicate versions
-        if len(all_data) > 0:
-            all_data = deduplicate(all_data)
-        else:
-            return None
+        # Search for latest version
+        stix_obj = latest_ver = None
+        for obj in all_data:
+            ver = obj.get("modified") or obj.get("created")
 
-        # reduce to most recent version
-        stix_obj = sorted(all_data, key=lambda k: k['modified'], reverse=True)[0]
+            if stix_obj is None or ver is None or ver > latest_ver:
+                stix_obj = obj
+                latest_ver = ver
 
         return stix_obj
 
--- a/stix2/test/v20/test_utils.py
+++ b/stix2/test/v20/test_utils.py
@@ -114,7 +114,7 @@ def test_deduplicate(stix_objs1):
     mods = [obj['modified'] for obj in unique]
 
     assert "indicator--00000000-0000-4000-8000-000000000001" in ids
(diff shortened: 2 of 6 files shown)


clenk:
Travis is failing: https://travis-ci.org/github/oasis-open/cti-python-stix2/jobs/719321125

Not sure why it's not showing up here on GitHub anymore.

chisholm:
I'm pretty sure the pattern_visitor.py and test_pattern_expression.py stuff is fixed in the pattern equivalence PR.  If I fixed it here, there would be merge conflicts there.  I fixed the missing comma in conftest.py though.

maybe-sybr:
#449 looks pretty big - is the expectation that it's good to go in now that it's up or would it be better to pull the lint fixes out and get master stable separately so this and other PRs can go in?

emmanvg:
@maybe-sybr I will rebase from master once the pre-commit fixes make it. We definitely want to make sure we are covering new code and tests.

clenk (approved):

PR #449, opened 2020-08-27
Graph Equivalence

emmanvg:
This PR provides the ability for a user to perform equivalence checks for two graphs.

- Moves semantic equivalence related functions to their own module. Added extra imports in `environment.py` to prevent breaking change
- Available via `stix2.environment.Environment.graphically_equivalent()` but also via `stix2.equivalence.graph_equivalence.graphically_equivalent()`
- Optional ability to perform versioned check when the are multiple versions of the same object present on the provided graph
- Configurable depth to prevent excessive recursion
- Ideally the complete graphs are provided - failure to de-reference object results in zero for the property contributing score
- More objects and property checks are available in Graph Based Semantic Equivalence


stix2/equivalence/graph_equivalence/__init__.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+# default weights used for the graph semantic equivalence process
+# values are re-balanced to account for new property checks and add up to 100
+WEIGHTS = {
+    "attack-pattern": {
+        "name": (30, partial_string_based),
+        "external_references": (60, partial_external_reference_based),
+        "created_by_ref": (5, semantic_check),
+        "object_marking_refs": (5, list_semantic_check),
```
clenk, line 102: As discussed offline, let's remove `created_by_ref` and `object_marking_refs` from the calculation since two organizations might create what is semantically the same content, or that same content might get marked differently by different organizations.

stix2/equivalence/object_equivalence/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+                    results[ref1] = {"matched": ref2, "value": result}
+    result = results.get(ref1, {}).get("value", 0.0)
+    logger.debug(
+        "--\t\t_versioned_checks '%s' '%s'\tresult: '%s'",
+        ref1, ref2, result,
+    )
+    return result
+
+
+def semantic_check(ref1, ref2, ds1, ds2, **weights):
```
clenk, line 323: It seems like this function is checking de-referenced objects rather than semantics in general. Maybe rename to `reference_check` or `dereferenced_check`? Same with `list_semantic_check`.

stix2/equivalence/object_equivalence/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+        "--\t\tsemantic_check '%s' '%s'\tresult: '%s'",
+        ref1, ref2, result,
+    )
+    return result
+
+
+def list_semantic_check(refs1, refs2, ds1, ds2, **weights):
+    """For objects that contain multiple references (i.e., object_refs) perform
+    the same de-reference for object-based semantic equivalence. The score influences
+    the objects containing all references. The result is weighted on the longest ref list"""
```
clenk, line 349: Missing a period.

clenk (commented):
Looking good!

Can we make the names of the semantic equivalence submodules consistent? Right now we have `graph_equivalence`, `object_equivalence`, and `patterns`. Since they're already under `equivalence` it's redundant to include it at this level too I think.

stix2/equivalence/graph/__init__.py
```diff
@@ -95,88 +100,36 @@ def graphically_equivalent(ds1, ds2, prop_scores={}, **weight_dict):
...
-        "sectors": (20, partial_list_based),
-    },
-    "indicator": {
-        "indicator_types": (15, partial_list_based),
-        "pattern": (80, custom_pattern_based),
-        "valid_from": (5, partial_timestamp_based),
-        "tdelta": 1,  # One day interval
-    },
-    "intrusion-set": {
+WEIGHTS.update({
```
clenk, line 103: Should we copy `WEIGHTS` and update/use the copy instead? That way if someone performs graph equivalence and then later performs object equivalence it won't do the additional stuff that graph equivalence does.

emmanvg: I think the fact that they also shared the same name caused a problem in scope resolution. I renamed the graph one...

clenk (approved):
LGTM, thanks @emmanvg!

PR #456, opened 2020-09-15
Fix object type tracking for AST comparison expression 'AND'

### LINKED ISSUE #455: AST comparison expression root type check for 'AND' is buggy
For example,

```
[(type-a:a=1 OR type-b:b=2) AND (type-a:a=2 OR type-b:b=1)]
```

is okay, but

```
[type-a:a=1 AND (type-a:a=2 OR type-b:b=1)]
```

is not:

```
>>> stix2.pattern_visitor.create_pattern_object("[type-a:a=1 AND (type-a:a=2 OR type-b:b=1)]", version="2.1")

<start of long stacktrace omitted>

  File "C:\Programming\python\stix\cti-python-stix2\stix2\patterns.py", line 534, in __init__
    super(AndBooleanExpression, self).__init__("AND", operands)
  File "C:\Programming\python\stix\cti-python-stix2\stix2\patterns.py", line 514, in __init__
    raise ValueError("All operands to an 'AND' expression must have the same object type")
ValueError: All operands to an 'AND' expression must have the same object type
```

The latter pattern simply drops a disjunct from the first 'OR' from the first pattern.  If the first is okay, the second should be too.  Another example is:

```
[(type-a:a=1 OR type-a:b=2) AND (type-a:a=2 OR type-b:b=1)]
```

This one simply changes the type of the second comparison in the first 'OR' to 'type-a', so that both types in that 'OR' are the same.  If the first example is okay, this one should be too.

chisholm:
Fixes #455 .

The spec is lacking in describing requirements on comparison expressions.  For AND, it says:

    The Comparison Expressions of an Observation Expression that use AND MUST use the same base Object Path, e.g., file:.

That sentence seems to assume that the operands to an AND are never sub-expressions.  If they are, it doesn't make sense to talk about their "base object path".  A sub-expression may be of arbitrary complexity, with lots of base object paths inside it.

The approach seems to be to make expressions which are syntactically valid but can never match, a spec violation.  In keeping with this idea, one might use it to better define requirements on comparison expression AND, e.g.:

    All operands of AND MUST be satisfiable with the same type of object.

The "satisfiable" idea is applicable to sub-expressions, and you can always tell what the applicable types are just by inspecting the pattern.  For OR, it's the union of all operand types.  For AND, it's the intersection.  For example:

    (type-a:a=1 OR type-b:b=2) AND (type-a:a=2 OR type-b:b=1)

The applicable types in the AND's lhs subexpression are {type-a, type-b}.  It's same on the right side.  Their intersection is {type-a, type-b}, which is non-empty, therefore the comparison expression is legal.  The AND operands are all satisfiable with the same type of object, and the possible types include type-a and type-b.  For:

    (type-a:a=1 OR type-b:b=2) AND (type-a:a=2 OR type-c:b=1)

We have {type-a, type-b} intersect {type-a, type-c}, which is {type-a}. So the only object type we can satisfy both AND operands with is type-a.  But since one exists, it's a legal pattern.

This is the idea implemented in this PR.  Instead of tracking a single root type, it tracks sets of types, and performs unions and intersections.  The approach is compatible with the spec and with existing library behavior and unit tests, and naturally extends to more complex comparison expressions.

### DIFF
--- a/stix2/patterns.py
+++ b/stix2/patterns.py
@@ -366,7 +366,7 @@ class _ComparisonExpression(_PatternExpression):
         else:
             self.rhs = make_constant(rhs)
         self.negated = negated
-        self.root_type = self.lhs.object_type_name
+        self.root_types = {self.lhs.object_type_name}
 
     def __str__(self):
         if self.negated:
@@ -506,15 +506,17 @@ class _BooleanExpression(_PatternExpression):
     """
     def __init__(self, operator, operands):
         self.operator = operator
-        self.operands = []
+        self.operands = list(operands)
         for arg in operands:
-            if not hasattr(self, "root_type"):
-                self.root_type = arg.root_type
-            elif self.root_type and (self.root_type != arg.root_type) and operator == "AND":
-                raise ValueError("All operands to an 'AND' expression must have the same object type")
-            elif self.root_type and (self.root_type != arg.root_type):
-                self.root_type = None
-            self.operands.append(arg)
+            if not hasattr(self, "root_types"):
+                self.root_types = arg.root_types
+            elif operator == "AND":
+                self.root_types &= arg.root_types
+            else:
+                self.root_types |= arg.root_types
+
+            if not self.root_types:
+                raise ValueError("All operands to an 'AND' expression must be satisfiable with the same object type")
 
     def __str__(self):
         sub_exprs = []
@@ -613,8 +615,8 @@ class ParentheticalExpression(_PatternExpression):
     """
     def __init__(self, exp):
         self.expression = exp
-        if hasattr(exp, "root_type"):
-            self.root_type = exp.root_type
+        if hasattr(exp, "root_types"):
+            self.root_types = exp.root_types
 
     def __str__(self):
         return "(%s)" % self.expression
--- a/stix2/test/v21/test_pattern_expressions.py
+++ b/stix2/test/v21/test_pattern_expressions.py
@@ -362,7 +362,7 @@ def test_parsing_or_observable_expression():
     assert str(exp) == "[user-account:account_type = 'unix' AND user-account:user_id = '1007' AND user-account:account_login = 'Peter'] OR [user-account:account_type = 'unix' AND user-account:user_id = '1008' AND user-account:account_login = 'Paul']"  # noqa
 
 
-def test_invalid_and_observable_expression():
+def test_invalid_and_comparison_expression():
     with pytest.raises(ValueError):
         stix2.AndBooleanExpression([
             stix2.EqualityComparisonExpression(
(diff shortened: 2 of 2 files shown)


rpiazza:
Interesting edge case, with elegant solution :-)

clenk (approved):

rpiazza:
@chisholm 

I just wanted to check that what I implemented in the elevator agrees with your understanding of the correct pattern expression semantics. 

Before the fix:

[ipv4-addr:value = '202.65.220.64'] OR [(domain-name:value = 'www.unog.dynssl.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64') OR (domain-name:value = 'www.unog.dnset.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64') OR (domain-name:value = 'www.unog.freetcp.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')] OR [mutex:name = '4htgsegvf'] OR [file:x_elevator_id = 'unog20120925' OR file:x_elevator_pivy_password = 'admin@338' OR file:hashes.MD5 = 'be6e72ad1b1ed2685a23dfe1b36f03cc']",

After the fix:

[ipv4-addr:value = '202.65.220.64' OR (domain-name:value = 'www.unog.dynssl.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64') OR (domain-name:value = 'www.unog.dnset.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64') OR mutex:name = '4htgsegvf' OR (domain-name:value = 'www.unog.freetcp.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64') OR file:x_elevator_id = 'unog20120925' OR file:x_elevator_pivy_password = 'admin@338' OR file:hashes.MD5 = 'be6e72ad1b1ed2685a23dfe1b36f03cc']",

The ORs seems to obviate the need of the CompoundObservationExpressions.

chisholm:
Those patterns are kind of hard to read, and I think github ate some of the characters (e.g. `domain-name:resolves_to_refs[].value` is not valid due to the empty brackets; I think there's supposed to be a "*" in there).  So I formatted them better to get a clearer idea of their structure:

```
[ipv4-addr:value = '202.65.220.64']
OR
[
  (domain-name:value = 'www.unog.dynssl.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')
  OR
  (domain-name:value = 'www.unog.dnset.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')
  OR
  (domain-name:value = 'www.unog.freetcp.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')
]
OR
[mutex:name = '4htgsegvf']
OR
[file:x_elevator_id = 'unog20120925' OR file:x_elevator_pivy_password = 'admin@338' OR file:hashes.MD5 = 'be6e72ad1b1ed2685a23dfe1b36f03cc']
```

```
[
  ipv4-addr:value = '202.65.220.64'
  OR
  (domain-name:value = 'www.unog.dynssl.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')
  OR
  (domain-name:value = 'www.unog.dnset.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')
  OR
  mutex:name = '4htgsegvf'
  OR
  (domain-name:value = 'www.unog.freetcp.com' AND domain-name:resolves_to_refs[*].value = '202.65.220.64')
  OR
  file:x_elevator_id = 'unog20120925'
  OR
  file:x_elevator_pivy_password = 'admin@338'
  OR
  file:hashes.MD5 = 'be6e72ad1b1ed2685a23dfe1b36f03cc'
]
```

Yes, I think they are the same.  In the first, you *can* bind only one observation; in the second, you *must* bind only one.

That shows a type of pattern normalization I hadn't thought of before: `[A] OR [B] OR ...` is equal to `[A OR B OR ...]`.  One could be converted to the other.  The semantic equivalence system wouldn't notice that equivalence, but maybe it should.  Which form is preferable?  It's an interesting observation (no pun intended :)).

PR #458, opened 2020-09-18
Versioning refinements

chisholm:
This intended to tighten up versionability detection and plug some holes.  It's not perfect, but hopefully an iterative improvement.

The changes include addressing versionability of a *type* separately from versionability of an *object* (i.e. an instance of a type).  This was kind of blurred before: STIX objects would be checked for presence of versioning properties; dicts would be checked for support from their object type.  This presents opportunities for bugs and misleading error messages.

For example, if a custom STIX type were created whose versioning properties were optional, and you didn't provide sufficient properties on an object, you'd get an error that you "cannot create new version of object of this type".  That's actually incorrect: you *can* creation new versions of instances of the type, you just have to provide sufficient properties on the object.  Also, if you tried to version a dict of a versionable type, didn't provide an explicit "modified" property for the new version as a kwarg, and it was also missing from the dict, you would get a crash because code just assumed the "modified" property would be in the dict.  In the first case, versionability of the object was checked but not the type; in the second case, versionability of the type was checked, but not the object.  So both types of error are made.

There is also the question of what to do if someone tries to version a dict of an unregistered STIX type.  Without registration, we can't tell what properties the type supports, so we can't determine its versionability.  Do we allow versioning?

Behavior included with this PR:
- If an object has all versioning properties, it is considered versionable.  Versionability of the type is not checked in this case.  This allows versioning properties to be added as custom properties as necessary, to enable versioning on an otherwise unversionable object.
- Versionability determination of a STIX type is the same as before.  All versioning properties must be supported.
- If a STIX type has not been registered and one attempts to version a dict for that type, a lax approach is taken, and the type will be considered versionable.  (This does not mean a particular instance of the type is versionable.)
- Except for the first bullet item, versionability determination of an object is only done if the type has been deemed versionable.  An object of versionable type is considered versionable if it has the "created" property.  If it does not have a "modified" property, it is assumed equal to "created", and the new version will have a "modified" property added with a timestamp greater than the created property.
- Exception types have been added to be clearer about why versioning fails.  You should no longer get an error saying a type is not versionable, when it is.  You should also no longer get a generic KeyError crash if a versionable instance of a versionable type does not have a "modified" property, and one was not given as a kwarg to new_version().

Possible future work: what if neither the type nor instance is considered versionable according to the above criteria, but *together*, the support/presence criteria are met for versionability?  E.g. type supports modified/revoked but not created, and instance has a custom "created" property but has neither "modified" nor "revoked".  The type would not be considered versionable, so versionability of the object would not even be considered.  But presence of the object's custom "created" property together with the type's support for "modified" and "revoked" put all the ingredients in place for versioning.

### DIFF
--- a/stix2/versioning.py
+++ b/stix2/versioning.py
@@ -124,7 +124,7 @@ def _is_versionable_type(data):
             # map to a registered class, and from that get a more complete
             # picture of its properties.
 
-            cls = stix2.registry.class_for_type(data["type"], stix_version)
+            cls = stix2.registry.class_for_type(data.get("type"), stix_version)
             if cls:
                 is_versionable = _VERSIONING_PROPERTIES.issubset(
                     cls._properties,
--- a/stix2/versioning.py
+++ b/stix2/versioning.py
@@ -9,13 +9,13 @@ import uuid
 import stix2.base
 import stix2.registry
 from stix2.utils import (
-    detect_spec_version, get_timestamp, is_sco, is_sdo, is_sro,
-    parse_into_datetime,
+    detect_spec_version, get_timestamp, is_sco, parse_into_datetime,
 )
 import stix2.v20
 
 from .exceptions import (
-    InvalidValueError, RevokeError, UnmodifiablePropertyError,
+    InvalidValueError, ObjectNotVersionableError, RevokeError,
+    TypeNotVersionableError, UnmodifiablePropertyError,
 )
 
 # STIX object properties that cannot be modified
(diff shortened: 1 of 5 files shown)


chisholm:
I checked the spec (e.g. sections 3.2 and 11.2), but couldn't find anything that put an all-or-nothing requirement on the usage of versioning properties for custom STIX types.  E.g. if you use "revoked", you must also use "created" and "modified".  It does say "The revoked property is only used by STIX Objects that support versioning...".  There is similar description for "modified", but not for "created".  That implies a policy which was used by spec authors for spec-defined objects.  There is not a "must" or "should" directing creators of custom types to add the other two properties.

It would be an odd situation to have a "revoked" property on an object/type you can never revoke, because it's unversionable.  I also don't think you're allowed to use properties named the same as common properties for purposes other than what the spec describes(?).  So, odd situation but it seems like the spec allows it.

Btw, I edited the PR description (updated the behavior bullets).  I hadn't quite got the correct behavioral description... maybe that is an indication of how tricky it is to get the logic right :)

stix2/versioning.py
```diff
@@ -88,36 +77,112 @@ def _is_versionable(data):
...
+            is_versionable = _VERSIONING_PROPERTIES.issubset(
+                data._properties,
+            )
+
+        elif isinstance(data, dict):
+            # Tougher to handle dicts.  We need to consider STIX version,
+            # map to a registered class, and from that get a more complete
+            # picture of its properties.
+
+            cls = stix2.registry.class_for_type(data["type"], stix_version)
```
clenk, line 127:
```suggestion
            cls = stix2.registry.class_for_type(data.get("type"), stix_version)
```

clenk (approved):

PR #464, opened 2020-10-15
Location objects now check for latitude and longitude before doing semantic equivalence

zrush-mitre:
Closes #463


clenk:
Superseded by #467.

PR #468, opened 2020-11-11
Extensions Support

emmanvg:
Currently in development...

Proposed Changes:
- Defines the STIXExtension Meta Object.
- Down on the basic use case (parsing, serializing) sometimes a chicken-and-egg situation can occur were an object being parsed requires an extension registered with the library. To prevent this problem without `allow_custom=True`, the library would check for the existence of an extension.
- The registration mechanism has an extra argument that would facilitate two registrations when a stix-extension is involved to define a new SDO or new SCO.
- More flexibility on the Extensions Map. This map now encapsulates extensions of the -ext kind and the stix-extension--<UUID> kind.


stix2/v20/observables.py
```diff
@@ -720,11 +720,11 @@ class WindowsRegistryKey(_Observable):
         ('modified', TimestampProperty()),
         ('creator_user_ref', ObjectReferenceProperty(valid_types='user-account')),
         ('number_of_subkeys', IntegerProperty()),
-        ('extensions', ExtensionsProperty(spec_version="2.0", enclosing_type=_type)),
+        ('extensions', ExtensionsProperty(spec_version='2.0')),
     ])
 
 
-class X509V3ExtenstionsType(_STIXBase20):
+class X509V3ExtensionsType(_STIXBase20):
```
clenk, line 727: This typo fix is great! But could we pull it out into a separate PR? In my mind it's not directly related to the rest of the Extensions support.

emmanvg:
Includes the following additional changes:
- Implements CS02 Incident Object
- Restriction changes to MarkingDefinition Object
- Updated all Object URLs to point at CS02

emmanvg:
- Removed imports that were kept for backwards compatibility
- Remove pretty=True from __str__() in _STIXBase
- Update docs to reflect the pretty change
- Added tests for ExtensionDefinition (simple object creation/serialization)
- Add new guide for Extensions

emmanvg:
Docs: https://stix2.readthedocs.io/en/dev-extensions-proposal/

rpiazza:
Custom extensions (section 11.3) are different than the new Extensions (section 7.3).  Custom extensions don't have and id as its key or an extension-type property.  I wasn't sure if you were making a distinction in the code.

chisholm:
Yes, extension_type can be omitted when registering an extension (which is how spec built-in extensions are too).  The CustomExtension decorator's `type` param can be an extension-definition STIX ID, for that type of extension, or a non STIX ID value for the old-style and builtin extensions.  So the library supports custom old style section 11, new style section 7.3, and spec builtin extensions.  That's the goal anyway.

rpiazza:
I was thinking that the test should have been a different defined property in the dict, i.e., strprop.  Otherwise this could be a test of having duplicate properties...

PR #474, opened 2020-11-19
fix ordering problem with Class definitions

### LINKED ISSUE #472: STIX Objects: Property ordering
Some object definitions do not contain their properties in the correct order. Affects the "pretty" serialization.

emmanvg:
closes #472


clenk (approved):
After looking through the examples in the spec, this looks like this lines up with the ordering used in those examples. :+1:

PR #480, opened 2020-12-22
Support Pagination in TAXII DataStore

### LINKED ISSUE #479: TAXIICollectionSource.query() only returns the first page
I noticed that the TAXIICollectionSource methods query(), relationships(), ... only operates with the first page returned by the associated TAXII 2.0 server.

I did tests with a local Medallion server v1.1.0 which have a default max_page_size of 100.

By calling TAXIICollectionSource.query() I can get only 100/10000 objects. Also, when calling relationships(stix2_id) the results are limited to those 100 objects.

The only solution I found is increasing that max_page_size, so all objects are returned in one call.

emmanvg:
Proposed solution to solve the pagination issues.

Closes #479

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ setup(
         'Bug Tracker': 'https://github.com/oasis-open/cti-python-stix2/issues/',
     },
     extras_require={
-        'taxii': ['taxii2-client>=2.2.1'],
+        'taxii': ['taxii2-client>=2.3.0'],
         'semantic': ['haversine', 'rapidfuzz'],
     },
 )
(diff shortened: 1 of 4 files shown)


chisholm:
Just curious: would it make sense to add an as_pages() function for the 2.1 TAXII client, similar to the 2.0 client?  It could be a nice convenience for 2.1 users and better align their APIs.

emmanvg:
I think so, that was one of my observations when making the changes. The second one being that some TAXII 2.0 Servers that claim pagination support are not providing the headers required to continue asking for pages making the client code to not behave with consistency.

clenk (approved):
LGTM; thanks @emmanvg!

PR #489, opened 2021-02-04
Fix observation expression DNF transformer to preserve FOLLOWEDBY order.

### LINKED ISSUE #488: Observation expression DNF transformer needs to preserve FOLLOWEDBY order
This was just a plain oversight on my part.  The same basic code handled both distributing AND over OR and FOLLOWEDBY over OR, but those have differing requirements.  FOLLOWEDBY order must always be preserved.  For example:

```
([A] OR [B]) FOLLOWEDBY ([C] AND [D])
```

was being transformed to

```
(([C] AND [D]) FOLLOWEDBY [A]) OR (([C] AND [D]) FOLLOWEDBY [B])
```

so that the A/B parts were put last in each distributed FOLLOWEDBY clause.  This would be fine if the FOLLOWEDBY were an AND.  The transformed result needs to be:

```
([A] FOLLOWEDBY ([C] AND [D])) OR ([B] FOLLOWEDBY ([C] AND [D]))
```

The transform needs to be fixed to preserve FOLLOWEDBY order.

chisholm:
Fixes #488 .

The algorithm is fixed to always preserve order, even with AND.  It is unnecessary for AND, but doesn't hurt.

### DIFF
--- a/stix2/equivalence/pattern/transform/observation.py
+++ b/stix2/equivalence/pattern/transform/observation.py
@@ -282,6 +282,7 @@ class AbsorptionTransformer(
 
         A or (A and B) = A
         A or (A followedby B) = A
+        A or (B followedby A) = A
 
     Other variants do not hold for observation expressions.
     """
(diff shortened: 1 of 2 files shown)


clenk (approved):
Thanks, @chisholm!

PR #492, opened 2021-02-18
Drop 'six' dependency (backwards breaking)

clenk:
We've already removed Python 2 from our test harness, but this removes all python 2 compatibility code that we left in.

Since we no longer need our Python 2 compatibility changes to the JSON canonicalization code, I reverted it to the version of the code at https://github.com/cyberphone/json-canonicalization/tree/master/python3. Since it follows a different code style than we do I excluded it from our pre-commit checks.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -10,7 +10,6 @@ known_third_party =
     pytz,
     requests,
     simplejson,
-    six,
     sphinx,
     stix2patterns,
     taxii2client,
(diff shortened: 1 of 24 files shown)


emmanvg:
wondering if the should also add a `python_requires=">=3.6"` in `setup.py` https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

PR #496, opened 2021-03-01
Similarity/Equivalence Changes

emmanvg:
- Expose algorithm configuration right on the calls
- Provide ability in object similarity/equivalence to de-reference objects via DataStores, or equality check when not provided
- Update documentation and guide


stix2/test/v21/test_environment.py
```diff
@@ -861,7 +858,9 @@ def test_object_similarity_exact_match():
 def test_non_existent_config_for_object():
```
clenk, line 858: should this test be renamed?

stix2/equivalence/object/__init__.py
```diff
@@ -439,6 +498,15 @@ def list_reference_check(refs1, refs2, ds1, ds2, **weights):
     return result
 
 
+def _datastore_check(ds1, ds2):
+    if (
+        issubclass(ds1.__class__, (DataStoreMixin, DataSink, DataSource)) or
+        issubclass(ds2.__class__, (DataStoreMixin, DataSink, DataSource))
```
clenk, line 504: I don't think DataSink should be included because you can't `.get()` from a DataSink. Sinks are designed just for adding, not querying.

stix2/environment.py
```diff
@@ -189,16 +189,31 @@ def creator_of(self, obj):
...
 
         Args:
             obj1: A stix2 object instance
             obj2: A stix2 object instance
             prop_scores: A dictionary that can hold individual property scores,
                 weights, contributing score, matching score and sum of weights.
-            weight_dict: A dictionary that can be used to override settings
-                in the similarity process
+            ds1: A DataStore object instance representing your graph
+            ds2: A DataStore object instance representing your graph
```
clenk, line 205: I think these two parameters could be explained better, and noted as optional. Maybe:

```suggestion
            ds1 (optional): A DataStore object instance from which to pull related objects
            ds2 (optional): A DataStore object instance from which to pull related objects
```

We could also add a note to the docstring (outside the Args section) about what will happen when ds1, ds2 are provided vs. when they won't.

Any changes to this docstring should then also be propogated to the other docstrings for consistency.

stix2/environment.py
```diff
@@ -253,17 +286,23 @@ def object_equivalence(obj1, obj2, prop_scores={}, threshold=70, **weight_dict):
...
-        return object_equivalence(obj1, obj2, prop_scores, threshold, **weight_dict)
+        return object_equivalence(
+            obj1, obj2, prop_scores, threshold, ds1, ds2,
+            ignore_spec_version, versioning_checks, max_depth, **weight_dict
+        )
 
     @staticmethod
-    def graph_similarity(ds1, ds2, prop_scores={}, **weight_dict):
+    def graph_similarity(
+        ds1, ds2, prop_scores={}, ignore_spec_version=False,
```
clenk, line 303: I think ds1, ds2 should default to `None` here, too, right?

stix2/environment.py
```diff
@@ -291,17 +339,24 @@ def graph_similarity(ds1, ds2, prop_scores={}, **weight_dict):
...
-        return graph_similarity(ds1, ds2, prop_scores, **weight_dict)
+        return graph_similarity(
+            ds1, ds2, prop_scores, ignore_spec_version,
+            versioning_checks, max_depth, **weight_dict
+        )
 
     @staticmethod
-    def graph_equivalence(ds1, ds2, prop_scores={}, threshold=70, **weight_dict):
+    def graph_equivalence(
+        ds1, ds2, prop_scores={}, threshold=70,
```
clenk, line 356: I think ds1, ds2 should default to `None` here, too, right?

clenk (changes requested):
Looks great for the most part; just a couple comments.

stix2/environment.py
```diff
@@ -236,8 +258,19 @@ def object_equivalence(obj1, obj2, prop_scores={}, threshold=70, **weight_dict):
             threshold: A numerical value between 0 and 100 to determine the minimum
                 score to result in successfully calling both objects equivalent. This
                 value can be tuned.
-            weight_dict: A dictionary that can be used to override settings
-                in the similarity process
+            ds1: A DataStore object instance representing your graph
+            ds2: A DataStore object instance representing your graph
```
emmanvg, line 262:
```suggestion
            ds1 (optional): A DataStore object instance from which to pull related objects
            ds2 (optional): A DataStore object instance from which to pull related objects
```

stix2/environment.py
emmanvg: I don't think we want to make them optional for the graph_* calls, otherwise there is nothing to test.

stix2/environment.py
emmanvg: I don't think we want to make them optional for the graph_* calls, otherwise there is nothing to test.

stix2/equivalence/object/__init__.py
emmanvg: removed DataSink in this check

clenk (approved):
Thank you @emmanvg!

PR #503, opened 2021-04-09
Revamp how STIX content customization is detected and enforced

### LINKED ISSUE #333: A Custom Object in object_refs of a Report Object.

### LINKED ISSUE #433: Cannot create custom references via `ObjectReferenceProperty`
`ObjectReferenceProperty` has an `allow_custom` field (and sets the attribute inside of `__init__()`, but does nothing with it in `clean()`)

### LINKED ISSUE #498: allow_custom is not inherited to common properties
The issue is that the allow_custom=True is not inherited in common properties that contain custom_properties.
Common STIX properties allow for adding custom properties using the allow_custom=True option. So, STIX does not restrict us for adding custom properties in a Common STIX property. 

An example to reproduce the issue is:
1. Create an "External Reference" common property and  add some custom properties
2. Create a Malware SDO and add the external-reference of the previous step.
3. Save the Malware SDO to a json file
3. Try to parse this Malware SDO via obj = parse(file_handle, allow_custom=True)

The error that is returned is the following:
"stix2.exceptions.InvalidValueError: Invalid value for Malware 'external_references': Unexpected properties for ExternalReference: (customizedproperty)."

This might be related to the issue #176 and issue #333. 
I apologize if it is a false positive and I am doing something wrong. Please let me know what I should do.

The json file is the following:

```{
    "type": "malware",
    "spec_version": "2.1",
    "id": "malware--83687fef-12ec-4959-94fe-a4f1c1380747",
    "created": "2021-03-06T18:54:28.533Z",
    "modified": "2021-03-06T18:54:28.533Z",
    "name": "malware-test-inheritance-of-custom-property",
    "description": "test-inheritance-of-custom-property",
    "is_family": false,
    "first_seen": "2021-03-06T18:54:28.533Z",
    "external_references": [
        {
            "source_name": "test-inheritance-custom",
            "description": "test for inheritance",
            "customizedproperty": "acustomproperty"
        }
    ]
}
```

### LINKED ISSUE #501: Note SDOs do not work on custom STIX types
As a user of cti-python-stix2, I want to be able to work with custom STIX types with the same level of functionality as I can with stock STIX types.

Currently, the `stix2` library throws an exception when loading STIX bundles containing a note referencing a custom STIX type. See the example below.

### Example STIX Bundle:

The following bundle contains a single [x-mitre-tactic](https://github.com/mitre/cti/blob/master/USAGE.md#tactics) object and an attached note.

```json
{
    "type": "bundle",
    "id": "bundle--ad5f3bce-004b-417e-899d-392f8591ab55",
    "spec_version": "2.1",
    "objects": [
        {
            "id": "x-mitre-tactic--d108ce10-2419-4cf9-a774-46161d6c6cfe",
            "created_by_ref": "identity--c78cb6e5-0c4b-4611-8297-d1b8b55e40b5",
            "name": "Collection",
            "description": "The adversary is trying to gather data of interest to their goal.\n\nCollection consists of techniques adversaries may use to gather information and the sources information is collected from that are relevant to following through on the adversary's objectives. Frequently, the next goal after collecting data is to steal (exfiltrate) the data. Common target sources include various drive types, browsers, audio, video, and email. Common collection methods include capturing screenshots and keyboard input.",
            "external_references": [
                {
                    "external_id": "TA0009",
                    "url": "https://attack.mitre.org/tactics/TA0009",
                    "source_name": "mitre-attack"
                }
            ],
            "object_marking_refs": [
                "marking-definition--fa42a846-8d90-4e51-bc29-71d5b4802168"
            ],
            "x_mitre_shortname": "collection",
            "type": "x-mitre-tactic",
            "modified": "2019-07-19T17:44:53.176Z",
            "created": "2018-10-17T00:14:20.652Z",
            "spec_version": "2.1",
            "x_mitre_domains": [
                "enterprise-attack"
            ],
            "x_mitre_version": "1.0"
        },
        {
            "type": "note",
            "id": "note--559fb05e-f4fc-4f93-9caa-e76f6aeb0207",
            "spec_version": "2.1",
            "abstract": "Tactic note example",
            "content": "Velit non dolore dolor. Quaerat quisquam dolor sit ipsum modi modi. Magnam porro voluptatem etincidunt adipisci sit tempora ut. Amet consectetur ipsum sit eius etincidunt. Dolorem porro consectetur tempora quaerat aliquam tempora dolore. Ut sit neque eius. Consectetur quisquam etincidunt dolor tempora dolore.",
            "object_refs": [
                "x-mitre-tactic--d108ce10-2419-4cf9-a774-46161d6c6cfe"
            ]
        }
    ]
}
```
### Example error on loading that bundle:
```
$ python3
>>> from stix2 import MemoryStore
>>> src = MemoryStore()
>>> src.load_from_file("tactic-note-bundle.json")
  File "/usr/local/lib/python3.8/site-packages/stix2/base.py", line 104, in _check_property
    kwargs[prop_name] = prop.clean(kwargs[prop_name])
  File "/usr/local/lib/python3.8/site-packages/stix2/properties.py", line 234, in clean
    result = [
  File "/usr/local/lib/python3.8/site-packages/stix2/properties.py", line 235, in <listcomp>
    self.contained.clean(item)
  File "/usr/local/lib/python3.8/site-packages/stix2/properties.py", line 509, in clean
    raise ValueError("The type-specifying prefix '%s' for this property is not valid" % (possible_prefix))
ValueError: The type-specifying prefix 'x-mitre-tactic' for this property is not valid

The above exception was the direct cause of the following exception:

  File "/usr/local/lib/python3.8/site-packages/stix2/datastore/memory.py", line 146, in load_from_file
    return self.source.load_from_file(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/stix2/datastore/memory.py", line 362, in load_from_file
    _add(self, stix_data, self.allow_custom, version)
  File "/usr/local/lib/python3.8/site-packages/stix2/datastore/memory.py", line 40, in _add
    _add(store, stix_obj, allow_custom, version)
  File "/usr/local/lib/python3.8/site-packages/stix2/datastore/memory.py", line 47, in _add
    stix_obj = parse(stix_data, allow_custom, version)
  File "/usr/local/lib/python3.8/site-packages/stix2/parsing.py", line 47, in parse
    obj = dict_to_stix2(obj, allow_custom, version)
  File "/usr/local/lib/python3.8/site-packages/stix2/parsing.py", line 142, in dict_to_stix2
    return obj_class(allow_custom=allow_custom, **stix_dict)
  File "/usr/local/lib/python3.8/site-packages/stix2/base.py", line 198, in __init__
  File "/usr/local/lib/python3.8/site-packages/stix2/base.py", line 110, in _check_property
    six.raise_from(
  File "<string>", line 3, in raise_from
stix2.exceptions.InvalidValueError: Invalid value for Note 'object_refs': The type-specifying prefix 'x-mitre-tactic' for this property is not valid
```

chisholm:
This PR shows work done to improve the way customization is handled.  One of the main changes is that STIX objects no longer maintain an internal `_allow_custom` flag.  In order to enforce that there is no customization (when allow_custom=False), one needs to know if customization is present.  Just because customizations were deemed okay at construction time (allow_custom=True) does not mean they actually exist.  So instead, objects are given a `has_custom` property, which directly addresses the question and makes customization detection in sub-objects possible.

The responsibility for both enforcing customization preferences and detecting the presence of customizations lies with the property `clean()` methods.  Before, those methods could not easily do their enforcement job because they were not told what the allow_custom preference was.  There was some very ugly hackage in places where that became a problem, which is now removed.  In this PR, the clean() methods are now passed the allow_custom preference.  They also now return not only the cleaned value, but also a boolean flag indicating whether customization was detected (which of course needs to be False if customization was not allowed!).

Enforcement has been tightened up in certain places.  This will catch customization in places where it didn't before, and in fact revealed some unit test bugs which have been fixed.  For example, open-vocabs are now properly enforced.  There is a new OpenVocabProperty class designed specifically for this, and STIX object classes have been changed to use it.  All of the enum and vocab values from the STIX spec have been added to special vocab modules for both 2.0 and 2.1; they are the enforcement criteria now.  The difference between an enum and open vocab is that no customization is ever allowed with enums, as that would result in non-compliant content; open vocabs admit values not listed in the spec, which are considered a customization and will be enforced accordingly.  Another example is that a customization in an extension will in all cases be caught and cause an error if allow_custom=False in the parent object.  The state of "customization" ought to always propagate upward.  An object has been customized if custom content occurs *anywhere* in the object.

Code dealing with hashes has been moved to a new module to remove clutter from the properties module.  Hash types are now properly enforced for each spec version, which are different in 2.0 and 2.1.  It supports some flexibility in how hash algorithms are named in user content, while ensuring that serialized output uses the spec-defined spelling/casing (which also differs between STIX 2.0 and 2.1).

Public API has mostly not changed, but user code which uses non-public APIs might break (e.g. if it called a clean() method).  Some property constructor args changed (e.g. allow_custom was removed since it doesn't make sense there).  Customization enforcement behavior has changed, so code/content which worked before might now trigger errors about custom content.

This PR includes changes which would address issue #501 .  The way in which ReferenceProperty does customization enforcement is a bit more sophisticated.  When a whitelist constraint on generic STIX type categories is given for this property and allow_custom=True during cleaning, the whitelist is automatically converted to a blacklist on the complementary categories.  This effectively weakens the criteria with respect to unregistered custom types: it ensures that types known to be incorrect are not given, as opposed to ensuring that types known to be correct are given.  An unregistered custom type is not known to be incorrect (the library knows nothing about it if it has not been registered), so it will satisfy the weakened criteria and be allowed.

Fixes #333.
Fixes #433.
Fixes #498.
Fixes #501.


stix2/v20/vocab.py
```diff
@@ -0,0 +1,178 @@
...
+
+TOOL_LABEL = [
+    "denial-of-service",
+    "exploitation",
+    "information-gathering",
+    "network-capture",
+    "credential-exploitation",
+    "remote-access",
+    "vulnerability-scanning",
+]
```
emmanvg, line 178: From an API perspective I think it would be more useful for users to have individual variables for all labels rather than making them figure out what index the correct value has...

``` python
from stix2.v20 import Tool
from stix2.v20.vocabs import TOOL_LABEL

t = Tool(labels=[TOOL_LABELS[0]]
```

vs

``` python
from stix2.v20 import Tool
from stix2.v20.vocabs import TOOL_LABEL_DENIAL_OF_SERVICE

t = Tool(labels=[TOOL_LABEL_DENIAL_OF_SERVICE]
```

```suggestion
TOOL_LABEL_DENIAL_OF_SERVICE =  "denial-of-service"
TOOL_LABEL_EXPLOITATION = "exploitation"
TOOL_LABEL_INFORMATION_GATHERING = "information-gathering"
TOOL_LABEL_NETWORK_CAPTURE = "network-capture"
TOOL_LABEL_CREDENTIAL_EXPLOITATION = "credential-exploitation"
TOOL_LABEL_REMOTE_ACCESS = "remote-access"
TOOL_LABEL_VULNERABILITY_SCANNING = "vulnerability-scanning"

TOOL_LABEL = [
    TOOL_LABEL_DENIAL_OF_SERVICE,
    TOOL_LABEL_EXPLOITATION,
    TOOL_LABEL_INFORMATION_GATHERING,
    TOOL_LABEL_NETWORK_CAPTURE,
    TOOL_LABEL_CREDENTIAL_EXPLOITATION,
    TOOL_LABEL_REMOTE_ACCESS,
    TOOL_LABEL_VULNERABILITY_SCANNING,
]
```

stix2/base.py
```diff
@@ -220,13 +247,16 @@ def __deepcopy__(self, memo):
         if isinstance(self, _Observable):
             # Assume: valid references in the original object are still valid in the new version
             new_inner['_valid_refs'] = {'*': '*'}
-        new_inner['allow_custom'] = self._allow_custom
-        return cls(**new_inner)
+        return cls(allow_custom=True, **new_inner)
```
emmanvg, line 250: Would this be?

```suggestion
        return cls(allow_custom=self._allow_custom, **new_inner)
```

clenk:
As we discussed elsewhere, let's change this PR to allow custom open vocab values again for now. In a future version we'll work out more granular customization settings to support checking them.

stix2/base.py
chisholm: Remember one of the key changes is that there is no longer an`_allow_custom` attribute maintained on stix2 objects.  So the suggestion can't work.  We could use `self.has_custom`, but it wouldn't do anything different.  If `self` has customizations, we want to allow them here; if `self` doesn't have customizations, allowing them doesn't matter (the implementation will notice there are no customizations and set the `has_custom` property to False).  So `allow_custom=True` is simple and gives us the correct behavior in both cases.  I guess one could argue though, that using `self.has_custom` *looks* better.

stix2/v20/vocab.py
chisholm: Well they wouldn't have to look up an index, they'd have to look up a variable name instead :)  I think the main benefit of this is more transparent code.  Just from reading the code, you couldn't tell which vocab value `TOOL_LABELS[0]` is.  But from `TOOL_LABEL_DENIAL_OF_SERVICE`, it's pretty obvious.  Another benefit is that if people use the variables instead of typing values themselves, they're less likely to make typos.

I'm willing to make variables if we want.

stix2/base.py
emmanvg: Ahh I see! No change here then.

stix2/v20/vocab.py
clenk: I agree; I think we should make change.

emmanvg:
The other general comment was to inspect the documentation/ipynb guides to check for any content that needs updating based on this changes.

adulau:
``` Customization enforcement behavior has changed, so code/content which worked before might now trigger errors about custom content.``` What's the background for such change in the reference implementation? Can you point me to the standard where this is a required change?

clenk:
@adulau the reason for these changes is not anything in the standard. There have been inconsistencies with this library's handling of custom content that has led to several issues, such as #501, #498, #433, #333. That's what we're trying to address here. But as part of that, some things have changed, such as the parameters and return values of the `clean()` method. Some content that gave custom content errors before in a nested object or reference properties are now allowed.

chisholm:
Having an `allow_custom` config setting was a software design decision, not a requirement from the specification.  It was thought to be desirable to allow users to cause the library to produce errors when custom content is encountered, when they want to prevent it.  It can also be useful for catching typos and mistakes.  That design decision is much older than this PR, and has been in the library since the beginning as far as I know.  This PR is about improving how that system works.

I mentioned in the PR summary that it could cause new errors, which sounds alarming; maybe I should also have been clearer that it can resolve errors too :)  The last paragraph describes one such case.

Issue #511, opened 2021-06-24
Properties ending in `_ref`/`_refs` should not be required to have a specific MRO


maybe-sybr:
At the moment, we check the MRO of a property instance used in a custom SCO definition in order to determine if it is valid for use as a property with a name ending in either `_ref` or `_refs`. The relevant code snippets are:

https://github.com/oasis-open/cti-python-stix2/blob/e9d417de2592c0c7367c312ca0fd25dc8f8a9818/stix2/registration.py#L119-L128
https://github.com/oasis-open/cti-python-stix2/blob/e9d417de2592c0c7367c312ca0fd25dc8f8a9818/stix2/utils.py#L305-L310

I'm hitting issues with this because I've created and am using an annotated reference type which is set up as a `DictionaryProperty` itselt containing a `ref` property. Since the `foo_ref` property I define is a subclass of a `DictionaryProperty` (and `bar_refs` would be a `ListProperty` containing `DictionaryProperty`s), the MRO check fails and the custom observable logic barfs on my definition. This is a problem for me since I've already used `_ref(s)` naming for some custom SDOs and to have to use something like `_refs_annotated` for observables would be weird and a pain in the butt.

The section of the spec which I think this logic has come from is the first two dot points in 3.1:
> Certain names of properties MUST have specific suffixes.
> * If the value of the property contains an ID reference for embedded relationships it MUST end in _ref
> * If the value of the property contains a list of embedded relationships it MUST end in _refs.

but IMO the implemeted logic is kind of backward. It states that properties containing (lists of) `identifier`(s) must be called something ending in `_ref(s)`, but doesn't say that anything ending in `_ref(s)` has to be a (list of) `identifier`(s) which is what we're checking. If my custom schema states that `foo_refs` is a `StringProperty` and that doesn't contradict the spec, I figure I should be able to codify that in the implementation.

I could accept an interpretation where any property name ending in `_ref(s)` must contain a (list of) `identifier`(s), and maybe ask the TC to clarify that in the spec, if it was acceptable to say that they must either be or contain an `identifier` at some well-defined sub-(sub-...)property.

[...]

PR #513, opened 2021-06-25
improv: Check reference prop types for all customs

maybe-sybr:
Previously the requirement that properties ending in `_ref(s)` were
instances of an appropriate type capable of containing a STIX
`identifier` (the current interpretation of section 3.1 of the spec) was
only enforced for custom observables. This change refactors the property
checks for custom objects to enforce this and the STIX 2.1 property name
requirement (also from section 3.1) in a common helper, therefore
extending the enforcement of both requirements to all custom object
types created by downstream projects.

ref: #511

### DIFF
--- a/stix2/registration.py
+++ b/stix2/registration.py
@@ -1,11 +1,71 @@
-import itertools
 import re
 
 from . import registry, version
 from .base import _DomainObject
 from .exceptions import DuplicateRegistrationError
-from .properties import _validate_type
-from .utils import PREFIX_21_REGEX, get_class_hierarchy_names
+from .properties import (
+    ListProperty, ObjectReferenceProperty, ReferenceProperty, _validate_type,
+)
+from .utils import PREFIX_21_REGEX
+
+
+def _validate_ref_props(props_map, is_observable20=False):
+    """
+    Validate that reference properties contain an expected type.
+
+    Properties ending in "_ref/s" need to be instances of specific types to
+    meet the specification requirements. For 2.0 and 2.1 conformance, these
+    properties are expected to be implemented with `ReferenceProperty` (or a
+    subclass thereof), except for the special case of STIX 2.0 observables
+    which must be implemented with `ObjectReferenceProperty`.
+
+    Args:
+        props_map (mapping): A mapping of STIX object properties to be checked.
+        is_observable20 (bool): Flag for the STIX 2.0 observables special case.
+
+    Raises:
+        ValueError: If the properties do not conform.
+    """
+    if is_observable20:
+        ref_prop_type = ObjectReferenceProperty
+    else:
+        ref_prop_type = ReferenceProperty
+    for prop_name, prop_obj in props_map.items():
+        tail = prop_name.rsplit("_", 1)[-1]
+        if tail == "ref" and not isinstance(prop_obj, ref_prop_type):
+            raise ValueError(
+                f"{prop_name!r} is named like a reference property but is not "
+                f"a subclass of {ref_prop_type.__name__!r}.",
+            )
+        elif tail == "refs" and not (
+            isinstance(prop_obj, ListProperty)
+            and isinstance(prop_obj.contained, ref_prop_type)
+        ):
+            raise ValueError(
+                f"{prop_name!r} is named like a reference list property but is not "
+                f"a 'ListProperty' containing a subclass of {ref_prop_type.__name__!r}.",
+            )
+
+
+def _validate_props(props_map, version, **kwargs):
+    """
+    Validate that a map of properties is conformant for this STIX `version`.
+
+    Args:
+        props_map (mapping): A mapping of STIX object properties to be checked.
+        version (str): Which STIX2 version the properties must confirm to.
+        kwargs (mapping): Arguments to pass on to specific property validators.
+
+    Raises:
+        ValueError: If the properties do not conform.
+    """
+    # Confirm conformance with STIX 2.1+ requirements for property names
+    if version != "2.0":
+        for prop_name, prop_value in props_map.items():
+            if not re.match(PREFIX_21_REGEX, prop_name):
+                raise ValueError("Property name '%s' must begin with an alpha character." % prop_name)
+    # Confirm conformance of reference properties
+    _validate_ref_props(props_map, **kwargs)
 
 
 def _register_object(new_type, version=version.DEFAULT_VERSION):
@@ -29,15 +89,10 @@ def _register_object(new_type, version=version.DEFAULT_VERSION):
             new_type.__name__,
         )
 
-    properties = new_type._properties
-
     if not version:
         version = version.DEFAULT_VERSION
 
-    if version == "2.1":
-        for prop_name, prop in properties.items():
-            if not re.match(PREFIX_21_REGEX, prop_name):
-                raise ValueError("Property name '%s' must begin with an alpha character" % prop_name)
+    _validate_props(new_type._properties, version)
 
     OBJ_MAP = registry.STIX2_OBJ_MAPS[version]['objects']
     if new_type._type in OBJ_MAP.keys():
@@ -54,19 +109,12 @@ def _register_marking(new_marking, version=version.DEFAULT_VERSION):
             None, use latest version.
 
     """
-
-    mark_type = new_marking._type
-    properties = new_marking._properties
-
     if not version:
         version = version.DEFAULT_VERSION
 
+    mark_type = new_marking._type
     _validate_type(mark_type, version)
-
-    if version == "2.1":
-        for prop_name, prop_value in properties.items():
-            if not re.match(PREFIX_21_REGEX, prop_name):
-                raise ValueError("Property name '%s' must begin with an alpha character." % prop_name)
+    _validate_props(new_marking._properties, version)
 
     OBJ_MAP_MARKING = registry.STIX2_OBJ_MAPS[version]['markings']
     if mark_type in OBJ_MAP_MARKING.keys():
@@ -83,49 +131,13 @@ def _register_observable(new_observable, version=version.DEFAULT_VERSION):
             None, use latest version.
 
     """
-    properties = new_observable._properties
-
     if not version:
         version = version.DEFAULT_VERSION
 
-    if version == "2.0":
-        # If using STIX2.0, check properties ending in "_ref/s" are ObjectReferenceProperties
-        for prop_name, prop in properties.items():
-            if prop_name.endswith('_ref') and ('ObjectReferenceProperty' not in get_class_hierarchy_names(prop)):
-                raise ValueError(
-                    "'%s' is named like an object reference property but "
-                    "is not an ObjectReferenceProperty." % prop_name,
-                )
-            elif (
-                prop_name.endswith('_refs') and (
-                    'ListProperty' not in get_class_hierarchy_names(prop) or
-                    'ObjectReferenceProperty' not in get_class_hierarchy_names(prop.contained)
-                )
-            ):
-                raise ValueError(
-                    "'%s' is named like an object reference list property but "
-                    "is not a ListProperty containing ObjectReferenceProperty." % prop_name,
-                )
-    else:
-        # If using STIX2.1 (or newer...), check properties ending in "_ref/s" are ReferenceProperties
-        for prop_name, prop in properties.items():
-            if not re.match(PREFIX_21_REGEX, prop_name):
-                raise ValueError("Property name '%s' must begin with an alpha character." % prop_name)
-            elif prop_name.endswith('_ref') and ('ReferenceProperty' not in get_class_hierarchy_names(prop)):
-                raise ValueError(
-                    "'%s' is named like a reference property but "
-                    "is not a ReferenceProperty." % prop_name,
-                )
-            elif (
-                prop_name.endswith('_refs') and (
-                    'ListProperty' not in get_class_hierarchy_names(prop) or
-                    'ReferenceProperty' not in get_class_hierarchy_names(prop.contained)
-                )
-            ):
-                raise ValueError(
-                    "'%s' is named like a reference list property but "
-                    "is not a ListProperty containing ReferenceProperty." % prop_name,
-                )
+    _validate_props(
+        new_observable._properties, version,
+        is_observable20=(version == "2.0"),
+    )
 
     OBJ_MAP_OBSERVABLE = registry.STIX2_OBJ_MAPS[version]['observables']
     if new_observable._type in OBJ_MAP_OBSERVABLE.keys():
@@ -146,20 +158,7 @@ def _register_extension(
     """
     ext_type = new_extension._type
 
-    # Need to check both toplevel and nested properties
-    prop_groups = [new_extension._properties]
-    if hasattr(new_extension, "_toplevel_properties"):
-        prop_groups.append(new_extension._toplevel_properties)
-    prop_names = itertools.chain.from_iterable(prop_groups)
-
     _validate_type(ext_type, version)
-
-    if not new_extension._properties:
-        raise ValueError(
-            "Invalid extension: must define at least one property: " +
-            ext_type,
-        )
-
     if version == "2.1":
         if not (ext_type.endswith('-ext') or ext_type.startswith('extension-definition--')):
             raise ValueError(
@@ -167,9 +166,23 @@ def _register_extension(
                 ext_type,
             )
 
-        for prop_name in prop_names:
-            if not re.match(PREFIX_21_REGEX, prop_name):
-                raise ValueError("Property name '%s' must begin with an alpha character." % prop_name)
+    tl_props = getattr(new_extension, "_toplevel_properties", None)
+    if any((
+        # There must always be at least one property in an extension. This
+        # holds for instances of both custom object extensions which must
+        # contain one or more custom properties, and `extension-definition`s
+        # which must contain an `extension_type` property.
+        not new_extension._properties,
+        # If a top-level properties mapping is provided, it cannot be empty
+        tl_props is not None and not tl_props,
+    )):
+        raise ValueError(
+            "Invalid extension: must define at least one property: " +
+            ext_type,
+        )
+    # We need to validate all properties related to this extension
+    combined_props = dict(new_extension._properties, **(tl_props or dict()))
+    _validate_props(combined_props, version)
 
     EXT_MAP = registry.STIX2_OBJ_MAPS[version]['extensions']
 
--- a/stix2/test/v20/test_custom.py
+++ b/stix2/test/v20/test_custom.py
@@ -385,6 +385,84 @@ def test_custom_object_invalid_type_name():
     assert "Invalid type name 'x_new_object':" in str(excinfo.value)
 
 
+def test_custom_object_ref_property_as_identifier():
+    @stix2.v20.CustomObject(
+        'x-new-obj-with-ref', [
+            ('property_ref', stix2.properties.ReferenceProperty(invalid_types=[])),
+        ],
+    )
+    class NewObs():
+        pass
+
+
+def test_custom_object_refs_property_containing_identifiers():
+    @stix2.v20.CustomObject(
+        'x-new-obj-with-refs', [
+            ('property_refs', stix2.properties.ListProperty(stix2.properties.ReferenceProperty(invalid_types=[]))),
+        ],
+    )
+    class NewObs():
+        pass
+
+
+def test_custom_object_ref_property_as_objectref():
+    with pytest.raises(ValueError, match=r"not a subclass of 'ReferenceProperty"):
+        @stix2.v20.CustomObject(
+            'x-new-obj-with-objref', [
+                ('property_ref', stix2.properties.ObjectReferenceProperty()),
+            ],
+        )
+        class NewObs():
+            pass
+
+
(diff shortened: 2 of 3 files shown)


stix2/registration.py
```diff
@@ -3,10 +3,83 @@
...
+from .properties import (
+    _validate_type, ListProperty, ObjectReferenceProperty, ReferenceProperty,
+)
 from .utils import PREFIX_21_REGEX, get_class_hierarchy_names
 from .version import DEFAULT_VERSION
 
+# Properties ending in "_ref/s" need to be instances of specific types to meet
+# our interpretation of section 3.1 of the spec
+_VERSION_REF_TYPES_MAP = {
+    # TODO: The 2.0 code seems to use `ReferenceProperty`s - is that okay?
```
maybe-sybr, line 15: > The 2.0 code seems to use `ReferenceProperty`s - is that okay?

rather than `ObjectReferenceProperty`s. AFAICT `ReferenceProperty` just seems to be a better version of the other? Is there some spec related reason why it's not valid in 2.0 documents?

stix2/registration.py
```diff
@@ -3,10 +3,83 @@
...
+        props_map (mapping): A mapping of STIX object properties to be checked.
+        version (str): Which STIX2 version the properties must confirm to.
+
+    Raises:
+        ValueError: If the properties do not conform.
+    """
+    # Confirm conformance with STIX 2.1 requirements for property names
+    #
+    # TODO: Should this be 2.1 or anything unknown assuming 2.1+, like we do
+    # for identifier properties above? `version != "2.0"`?
```
maybe-sybr, line 71: > TODO: Should this be 2.1 or anything unknown assuming 2.1+, like we do for identifier properties above? `version != "2.0"`?

stix2/registration.py
```diff
@@ -3,10 +3,83 @@
...
+    if version == "2.1":
+        for prop_name, prop_value in props_map.items():
+            # TODO: We could probably avoid doing a regex here just for the
+            # first char, but actually there are other constraints in 3.1 which
+            # we don't check which a regex would be suitable for
+            # (e.g. don't use hyphens in prop names, length min/max)
```
maybe-sybr, line 77: Just and interesting note for work that could be done here or in a future PR

maybe-sybr:
Lint should be gone now

stix2/registration.py
```diff
@@ -3,10 +3,83 @@
...
+            raise ValueError(
+                f"{prop_name!r} is named like a reference property but is not "
+                f"a subclass of any of {ref_prop_types!r}.",
+            )
+        elif tail == "refs" and not all((
+            isinstance(prop_obj, ListProperty),
+            isinstance(getattr(prop_obj, "contained", None), ref_prop_types),
+        )):
+            raise ValueError(
+                f"{prop_name!r} is named like a reference list property but is not"
```
maybe-sybr, line 52: Needs a trailing space in this string

stix2/registration.py
```diff
@@ -3,10 +3,83 @@
...
+from .utils import PREFIX_21_REGEX
 from .version import DEFAULT_VERSION
 
+# Properties ending in "_ref/s" need to be instances of specific types to meet
+# our interpretation of section 3.1 of the spec
+_VERSION_REF_TYPES_MAP = {
+    # TODO: The 2.0 code seems to use `ReferenceProperty`s - is that okay?
+    "2.0": (ObjectReferenceProperty, ReferenceProperty),
+    "2.1": (ReferenceProperty,),
+}
```
maybe-sybr, line 18: Would it be reasonable to make this public (remove the underscore) and turn these values from `tuple`s into `set`s (or something similarly mutable) so that a downstream developer like me could do something like `VERSION_REF_TYPES_MAP["2.1"].add(MyCustomReferenceType)`?

stix2/registration.py
```diff
@@ -3,10 +3,83 @@
...
+        ValueError: If the properties do not conform.
+    """
+    try:
+        ref_prop_types = _VERSION_REF_TYPES_MAP[version]
+    except KeyError:
+        ref_prop_types = _UNKNOWN_VERSION_REF_TYPES
+
+    for prop_name, prop_obj in props_map.items():
+        tail = prop_name.rsplit("_", 1)[-1]
+        if tail == "ref" and not isinstance(prop_obj, ref_prop_types):
```
maybe-sybr, line 42: Alternatively, if we turned `ReferenceProperty` (or all `Property`/`STIXBase` types?) use the `abc.ABCMeta` metaclass, it would be possible to do `ReferenceProperty.register(MyCustomReferenceType)` and have this subclass check pass for registered virtual subclasses.

That approach would introduce a new method on those STIX object types which could arguably collide with some STIX (custom or future) attribute, which probably isn't great. Alternatively we could implement a `ReferenceProperty.__subclasshook__()` where we codify what a property needs to look like to be considered valid for use as a property ending in `_ref(s)`. However, that would result in some non-obvious data validation hidden behind these `isinstance()` calls by being implemented in the `__subclasshook__()`. Not sure I particularly like it because of that.

maybe-sybr (commented):
Just some thoughts on how I might be able to solve my issue with an annotated reference property as described in #511 - these don't need to be resolved for this changeset to be mergable, but it makes sense to make the commentary where the `isinstance()` checks exist in the diff

stix2/registration.py
maybe-sybr: As a note for myself, `stix2.properties.Property` is not a subclass of `_STIXBase`

maybe-sybr:
I've rebased this on top of master now that #468 is in to resolve the conflict in `_register_extension`. Everything appears to still be behaving as expected.

maybe-sybr:
I've de-drafted this and fixed/removed the todos (the threads above can still be responded to though) to make life easier if you're satisfied with merging this prior to releasing 3.0.0.

I'd like to implement one of the options for adding extra valid reference types, and I have an example  of the ABCMeta approach on my staging branch at maybe-sybr/cti-python-stix2@a8fe790 , but if I have to keep a staging branch on top of 3.0.0 just for that one change, I could live with that :)

stix2/registration.py
maybe-sybr: I made this `!= "2.0"` in the current diff with the expectation that we can amend it later if we need. But LMK if you'd prefer the explicit `== "2.1"`.

maybe-sybr:
@rpiazza I've rebased this on top of v3.0.0 and resolved a conflict with the new top level property validation done for custom extensions. The tests pass but I'll make a quick comment on the diff for @chisholm to sanity check to make sure I didn't mess the expected behaviour.

stix2/registration.py
```diff
@@ -146,30 +159,25 @@ def _register_extension(
...
+    combined_props = {
+        **new_extension._properties,
+        **getattr(new_extension, "_toplevel_properties", dict()),
+    }
+    if not combined_props:
+        raise ValueError(
+            "Invalid extension: must define at least one property: " +
+            ext_type,
+        )
+    _validate_props(combined_props, version)
```
maybe-sybr, line 180: @chisholm - per my comment above, could you quickly look over this bit and let me know if I've correctly interpreted how my extra validation should function.

Notably, your change would still raise a `ValueError` if the `_properties` were empty but the `_top_level_properties` were not. I'm not entirely confident that my check against the `combined_properties` is correct or if a spec-complicant construction of a custom extension MUST have a property (ie. the extension type `top-level-property`) if any `_top_level_properties` are defined.

chisholm: `_properties` musn't be empty.  At the very least, you need `extension_type` as a nested property, which describes the type of extension.  Checking for the existence of that particular nested property could be an additional bit of validation I guess.

I suspect we don't really envision people calling `_register_extension()` directly, but if they did, then checking `_toplevel_properties` against `extension_type` might also make sense.  Or maybe toplevel properties are just ignored if extension_type is not "toplevel-property-extension".  Or maybe we produce a warning.  Anyway, if you use the `@CustomExtension` decorator, that should be done correctly for you.

stix2/registration.py
chisholm: I think you're suggesting having different ReferenceProperty subclasses for enforcing different ref property requirements?  I don't think the requirements on ref properties are complicated enough that we need specialized custom code for different cases.  The black/whitelist on STIX types and type classes seems sufficient.  Can you think of a case where that is not sufficient?

A stylistic note: I don't like using exception handling for non-exceptional circumstances, or if the same thing could be easily done in a different way.  Instead of the KeyError exception stuff, how about:

```python
ref_prop_types = _VERSION_REF_TYPES_MAP.get(version, _UNKNOWN_VERSION_REF_TYPES)
```

?

chisholm:
Overall, I think doing the ref property checking consistently makes sense.  It didn't make much sense to do it only for observables.  And I definitely like getting rid of the mro lookups!  I always thought those were ugly.  It's.... approximately correct I think, but not exactly.  E.g. if you registered a custom 2.0 SDO but used `ObjectReferenceProperty` for its ref properties, I think it would pass the checks.  But it shouldn't, since that property class is intended only for 2.0 SCO ref properties.

stix2/registration.py
maybe-sybr: > I think you're suggesting having different ReferenceProperty subclasses for enforcing different ref property requirements? I don't think the requirements on ref properties are complicated enough that we need specialized custom code for different cases. The black/whitelist on STIX types and type classes seems sufficient. Can you think of a case where that is not sufficient?
> I think you're suggesting having different ReferenceProperty subclasses for enforcing different ref property requirements? I don't think the requirements on ref properties are complicated enough that we need specialized custom code for different cases. The black/whitelist on STIX types and type classes seems sufficient. Can you think of a case where that is not sufficient?

The background in #511 which this change is in support of may be illustrative, in particular this https://github.com/oasis-open/cti-python-stix2/issues/511#issuecomment-868015761 . The tl;dr is that I have a thing which I consider a reference property (reference plus some "value-add") but which is a subclass of `DictionaryProperty` rather that `ReferenceProperty`. I could do some trickery with multiple inheritance and avoiding the use of `super()` but I really don't want to do that and would prefer to be able to register my `AnnotatedReference` as a virtual subclass using the ABCMeta machinery. This would be a change which is probably exclusively useful for supporting my exotic use case though, which is why I've not implemented that in this MR without us landing on a mutually acceptable approach :)

> Instead of the KeyError exception stuff, how about ...
Sure, will do.

stix2/registration.py
maybe-sybr: Yeah, I didn't mean to suggest that this would be called directly either - it's private so buyer beware IMO.

I think I'd like to keep this change from absorbing too much extra logic for validation if possible and just keep it to what's necessary to enable to checks for reference properties. I think to validate extension properties and potential top-level-properties to make sure they meet section 7.3 or section 11 requirements for custom extensions would need to live in this function anyway so it could be done in a separate MR. Perhaps we can take that on notice if you care to?

What I'm taking from your reply is that I should alter this logic to always ensure that `_properties` is not empty, rather than `combined_props`, which will restore the logic as it currently exists on v3.0.0/master. I'll still call `_validate_props(combined_props)` as the diff currently does, meaning that all property names and `ReferenceProperty` (and variant) type validations will still be performed as usual.

maybe-sybr:
> Overall, I think doing the ref property checking consistently makes sense. It didn't make much sense to do it only for observables. And I definitely like getting rid of the mro lookups! I always thought those were ugly.

Great, hopefully with the context around my use case for `ReferenceProperty` variants as well, my suggestions in https://github.com/oasis-open/cti-python-stix2/pull/513#pullrequestreview-697757608 will make some more sense. LMK what you think.

> It's.... approximately correct I think, but not exactly. E.g. if you registered a custom 2.0 SDO but used `ObjectReferenceProperty` for its ref properties, I think it would pass the checks. But it shouldn't, since that property class is intended only for 2.0 SCO ref properties.

I've just had a quick skim through the 2.0 spec (specifically part 3 and 4) and I think I understand how this has evolved now. I suppose the most correct thing would be to have a a nested mapping like `VALID_REF_TYPES[stix_version][stix_object_type]`, but that does sound pretty clunky for a single case. Alternatively, we could allow this logic to stand since I assume the 2.0 code will be relatively static in future, and the only risk is for downstream authors doing the "wrong thing". Happy to make any changes at your suggestion.

stix2/registration.py
chisholm: > I think I'd like to keep this change from absorbing too much extra logic for validation if possible and just keep it to what's necessary to enable to checks for reference properties. I think to validate extension properties and potential top-level-properties to make sure they meet section 7.3 or section 11 requirements for custom extensions would need to live in this function anyway so it could be done in a separate MR. Perhaps we can take that on notice if you care to?

The extra validation wasn't there before, so evidently no one has so far thought it necessary.  So you wouldn't have to include it with this.

> What I'm taking from your reply is that I should alter this logic to always ensure that `_properties` is not empty, rather than `combined_props`, which will restore the logic as it currently exists on v3.0.0/master. I'll still call `_validate_props(combined_props)` as the diff currently does, meaning that all property names and `ReferenceProperty` (and variant) type validations will still be performed as usual.

Yeah.  I looked to see if the spec had a requirement that a toplevel property extension must define at least one toplevel property, but I couldn't find that.

maybe-sybr: I actually just put that into the new diff since it seemed meaningless to have a top-level-property extension that didn't include anything. Check out the new diff and see what you think.

stix2/registration.py
```diff
@@ -146,30 +155,31 @@ def _register_extension(
...
         if not (ext_type.endswith('-ext') or ext_type.startswith('extension-definition--')):
             raise ValueError(
                 "Invalid extension type name '%s': must end with '-ext' or start with 'extension-definition--<UUID>'." %
                 ext_type,
             )
 
-        for prop_name in prop_names:
-            if not re.match(PREFIX_21_REGEX, prop_name):
-                raise ValueError("Property name '%s' must begin with an alpha character." % prop_name)
+    tl_props = getattr(new_extension, "_toplevel_properties")
```
maybe-sybr, line 166: This and the code below would be more elegant if we changed `custom.py` to set the `_CustomExtension._toplevel_properties` attribute to `None` rather than not defining it at all for non-top-level-property extensions.

stix2/registration.py
```diff
@@ -1,11 +1,71 @@
...
+    Validate that reference properties contain an expected type.
+
+    Args:
+        props_map (mapping): A mapping of STIX object properties to be checked.
+        version (str): Which STIX2 version the properties must confirm to.
+
+    Raises:
+        ValueError: If the properties do not conform.
+    """
+    ref_prop_types _VERSION_REF_TYPES_MAP.get(version, _UNKNOWN_VERSION_REF_TYPES)
```
maybe-sybr, line 33: oops!

maybe-sybr: fixed

maybe-sybr:
Just marking this as a draft while it needs autosquashing.

chisholm:
> > It's.... approximately correct I think, but not exactly. E.g. if you registered a custom 2.0 SDO but used `ObjectReferenceProperty` for its ref properties, I think it would pass the checks. But it shouldn't, since that property class is intended only for 2.0 SCO ref properties.
> 
> I've just had a quick skim through the 2.0 spec (specifically part 3 and 4) and I think I understand how this has evolved now. I suppose the most correct thing would be to have a a nested mapping like `VALID_REF_TYPES[stix_version][stix_object_type]`, but that does sound pretty clunky for a single case. Alternatively, we could allow this logic to stand since I assume the 2.0 code will be relatively static in future, and the only risk is for downstream authors doing the "wrong thing". Happy to make any changes at your suggestion.

Nah, I think treatment would only need to depend on the "class" of object, e.g. SDO vs SCO.  So it would only need to be (notionally) `VALID_REF_TYPES[stix_version][stix_type_class]`.  And maybe as simple as `VALID_REF_TYPES[2.0 SCOs]` and `VALID_REF_TYPES[everything else]`.

I don't know how important it is now to be fanatically correct w.r.t. STIX 2.0, but this *is* a reference implementation... :)

stix2/registration.py
chisholm: So instead you would have:

```python
tl_props = new_extension._toplevel_properties
```

And the code below would not change?  The change doesn't seem all that significant.  What did you have in mind?

maybe-sybr: Yep. It'd mostly just mean that all extensions have a `_top_level_properties` rather than having to do `getattr()`s on it. In my head, being able to always access an expected attribute and react when it's set to a value defining "not present on the represented object" feels nice. I'm not stressed about changing this if you're satisfied with the code as it stands though

chisholm:
If we're going to do this, let's just do it right.  It should be easy to adapt your code to:
- Check for `ObjectReferenceProperty` for 2.0 SCOs
- Check for `ReferenceProperty` for all other objects

Then ensure there are unit tests to:
- ensure a 2.0 SCO with `ReferenceProperty` errors
- ensure a 2.0 non-SCO with `ObjectReferenceProperty` errors
- ensure a 2.1 object errors with `ObjectReferenceProperty`

maybe-sybr:
Check out the new diff, @chisholm . I've decided against having a more complex mapping since the only special case we have to deal with is 2.0 observables and it's much more obvious to codify that with a single conditional in the reference property validation helper. I imagine that future iterations of the 2.X spec will aim to maintain coherence for all properties that appear to be references, so building infrastructure to support future deviations is just an exercise in complexity for its own sake. That should be fairly easy to pivot back to in future if the need arises (or in this PR if that's your preference, just LMK).

There are tests for custom objects and observables to ensure that reference vs object reference properties are accepted and rejected appropriately. I don't do any testing for acceptance of subclasses, LMK if you'd like to see that. Since supporting that was the point of #511, I don't mind putting it in, but since the use case is now more dubiously useful, I could also see it being valid to omit them.

stix2/registration.py
```diff
@@ -83,49 +132,13 @@ def _register_observable(new_observable, version=version.DEFAULT_VERSION):
...
-                )
-            ):
-                raise ValueError(
-                    "'%s' is named like a reference list property but "
-                    "is not a ListProperty containing ReferenceProperty." % prop_name,
-                )
+    _validate_props(
+        new_observable._properties, version,
+        is_observable20=(version == "2.0"),
+    )
```
maybe-sybr, line 141: Alternatively to this approach where `_validate_props()` is responsible for calling `_validate_ref_props()`, we could force each registration helper to call both of those methods themselves, which would  allow this specific one to pass in an alternate `ref_prop_type` kwarg to validate properties against. e.g.:

```python
_validate_props(new_observable._properties, version)
_validate_ref_props(new_observable._properties, version, ref_prop_type=ObjectReferenceProperty)
```

Doing so would introduce the possibility for drift in the validation steps applied between objects again though, which was the original problem with some requirements only being validated for custom observable objects. Not a deal breaker, just a risk to think about.

chisholm: "Reference property validation" is a subset of "property validation", so I like the idea of a general property validation function doing all property validation, as opposed to having special cases it does not validate.  I'm generally good with the design.

stix2/registration.py
```diff
@@ -1,11 +1,72 @@
...
 from .exceptions import DuplicateRegistrationError
-from .properties import _validate_type
-from .utils import PREFIX_21_REGEX, get_class_hierarchy_names
+from .properties import (
+    ListProperty, ObjectReferenceProperty, ReferenceProperty, _validate_type,
+)
+from .utils import PREFIX_21_REGEX
+
+
+def _validate_ref_props(props_map, version, is_observable20=False):
```
chisholm, line 12: `version` here is no longer used.  I think this param can be deleted.

stix2/registration.py
```diff
@@ -1,11 +1,72 @@
...
+        tail = prop_name.rsplit("_", 1)[-1]
+        if tail == "ref" and not isinstance(prop_obj, ref_prop_type):
+            raise ValueError(
+                f"{prop_name!r} is named like a reference property but is not "
+                f"a subclass of {ref_prop_type.__name__!r}.",
+            )
+        elif tail == "refs" and not all((
+            isinstance(prop_obj, ListProperty),
+            isinstance(getattr(prop_obj, "contained", None), ref_prop_type),
+        )):
```
chisholm, line 44: This use of `all()` undermines the normal short-circuiting behavior of boolean evaluation.  Here, all args must always be evaluated before `all()` itself can be called.  But you could take advantage of short-circuiting here to drop the `getattr()` and simplify the expression.  E.g.:

```python
        elif tail == "refs" and not (
            isinstance(prop_obj, ListProperty)
            and isinstance(prop_obj.contained, ref_prop_type)
        ):
```

Here, `prop_obj.contained` will not be evaluated unless the previous conjunct (`isinstance(prop_obj, ListProperty)`) evaluated to True.  Therefore, at the point of the attribute access, prop_obj must be a ListProperty and the access is guaranteed to work.  So you don't need `getattr()`.

stix2/registration.py
maybe-sybr: Ah yes, good catch

stix2/registration.py
maybe-sybr: Done in the new diff

stix2/registration.py
maybe-sybr: Done in the new diff

maybe-sybr: Done in the new diff

maybe-sybr:
Thanks @chisholm :)

PR #526, opened 2021-08-24
Fix TAXIICollectionSource.query() to propagate HTTPErrors

### LINKED ISSUE #524: stix2 3.0.0 - Filter object doesn’t work properly
For the previous version of stix2, this Filter returned the correct value of the Attack Pattern - 
`attack_pattern_name = tc_source.query([
            Filter("external_references.external_id", "=", attack_id),
            Filter("type", "=", "attack-pattern")
        ])[0]['name']`
For the new version, the Filter always returns the same value - `Code Signing Policy Modification`

The code:

collection_id = f"stix/collections/{ENTERPRISE_COLLECTION_ID}/"
collection_url = urljoin('https://cti-taxii.mitre.org', collection_id)
collection_data = Collection(collection_url, verify=client.verify, proxies=client.proxies)
tc_source = TAXIICollectionSource(collection_data)
attack_pattern_name = tc_source.query([
    Filter("external_references.external_id", "=", attack_id),
    Filter("type", "=", "attack-pattern")
])[0]['name']

### LINKED ISSUE #525: TAXIICollectionSource.query() hides and ignores all non-404 paging HTTP errors
The code is in a try-except block which catches all HTTPErrors but only does anything with 404s.  In general we should not silently ignore things like this.  The code is written such that when an error occurs and is ignored, important steps like deduplication and "local" filtering never happen (since they were located inside the try-except block), so the query results wind up being mysteriously incorrect.

In particular, I found a problem with [416 (Range Not Satisifiable)](https://www.rfc-editor.org/rfc/rfc7233.html#section-4.4).  That status code shouldn't necessarily be interpreted as an error at all.  It can naturally occur during TAXII 2.0 paging when you run past the end of the pages.  Running past the end of the pages can occur for example, whenever the server isn't sending `Content-Range` headers, so the only way for the client to know when it has hit the end is to run off the end and not get any results.  Currently when this happens, the error is silently ignored, the deduplication and local-filtering steps are skipped, and you get incorrect results.

chisholm:
This PR changes TAXIICollectionSource.query() to propagate HTTPErrors instead of ignoring them.  HTTP status code 416 is still ignored, since that can occur naturally while paging with TAXII 2.0.  The scope of the try-except is reduced to the minimum, to ensure steps are not skipped.

Fixes #524 
Fixes #525

### DIFF
--- a/stix2/datastore/taxii.py
+++ b/stix2/datastore/taxii.py
@@ -308,11 +308,13 @@ class TAXIICollectionSource(DataSource):
                     " the supplied TAXII Collection object are either not found or access is"
                     " denied. Received error: ", e,
                 )
+
+            # TAXII 2.0 paging can result in a 416 (Range Not Satisfiable) if
+            # the server isn't sending Content-Range headers, so the pager just
+            # goes until it runs out of pages.  So 416 can't be treated as a
+            # real error, just an end-of-pages condition.  For other codes,
+            # propagate the exception.
             elif e.response.status_code != 416:
-                # TAXII 2.0 paging can result in a 416 (Range Not Satisfiable)
-                # if the server isn't sending Content-Range headers, so the
-                # pager just goes until it runs out of pages.  So 416 can't be
-                # treated as a real error, just an end-of-pages condition.
                 raise
 
         # deduplicate data (before filtering as reduces wasted filtering)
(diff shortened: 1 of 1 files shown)


rpiazza:
Maybe it is me, but I missed the != 416, so it seemed like the comment contradicted the condition.

Maybe add "For other codes, propagate the exception"

chisholm:
Are you referring to a comment in the source code or this PR description?

chisholm:
I think you must have meant the code comment.  I moved the comment outside the "if" statement, and added your suggestion.

PR #534, opened 2021-12-23
Fix pattern semantic equivalence comparison expression DNF transformer bug

### LINKED ISSUE #533: Pattern semantic equivalence comparison expression DNF transformer ignores SCO types
The pattern semantic equivalence comparison expression DNF transformer rearranges an AST and can create new comparison expression AND nodes.  Current AST AND node implementation will check whether all operands have an SCO type in common, as required by spec, and error if this is not the case.  Some STIX patterns would cause the DNF transformer to create an AND node with no SCO types in common and trigger that error and crash.

For example:

    [(a:b=1 OR b:c=1) AND (b:d=1 OR c:d=1)]

would DNF transform to:

    [(a:b=1 AND b:d=1) OR (a:b=1 AND c:d=1) OR (b:c=1 AND b:d=1) OR (b:c=1 AND c:d=1)]

The first, second, and fourth AND expressions mix different SCO types and cause the error.

chisholm:
Fixes #533 .

This PR fixes the comparison expression DNF transformer to catch the exception caused by inconsistent SCO types among comparison expression AND operands, and drop the AND expression.  This resolves the problem and additionally enables a nice simplification.  For example, the example from the issue was:

    [(a:b=1 OR b:c=1) AND (b:d=1 OR c:d=1)]

which would have transformed to:

    [(a:b=1 AND b:d=1) OR (a:b=1 AND c:d=1) OR (b:c=1 AND b:d=1) OR (b:c=1 AND c:d=1)]

The transformer will now drop the first, second, and fourth AND expressions due to their mixed SCO types (which means they'd be impossible to satisfy anyway).  The result is simply:

    [b:c=1 AND b:d=1]

which is semantically equivalent to the original: only a `b` SCO could have satisfied it, which is obvious in its simplified form.

Also:
- Added a couple more unit tests to test this situation.  All of the original tests were designed to create different AST structures to exercise the transformers, and were not thought of in terms of mixing SCO types.  So none of the tests caught the error.
- Changed `_dupe_ast()` in the comparison expression transformer module to dupe simple \<path> \<op> \<value> comparison expressions.  Previously it did not do this because I think at the time, no transformation of that type of simple comparison expression happened.  So there was no need.  But special context-sensitive constant canonicalization was added, which does change that type of comparison expression (in particular, the constant portion).  So at the very least, the old comment about those expressions not changing was wrong.  The stated rationale was wrong.  I decided since the rationale was wrong now, I should probably go ahead and dupe them.
- The patterns.py change allowed the `_BooleanExpression` constructor to support an iterator for its `operands` parameter.  The old code implied the parameter value must support iteration over the values twice, which is not possible with an iterator.  It was a very minor change.

### DIFF
--- a/stix2/equivalence/pattern/transform/comparison.py
+++ b/stix2/equivalence/pattern/transform/comparison.py
@@ -13,20 +13,14 @@ from stix2.equivalence.pattern.transform.specials import (
     ipv4_addr, ipv6_addr, windows_reg_key,
 )
 from stix2.patterns import (
-    AndBooleanExpression, OrBooleanExpression, ParentheticalExpression,
-    _BooleanExpression, _ComparisonExpression,
+    AndBooleanExpression, ObjectPath, OrBooleanExpression,
+    ParentheticalExpression, _BooleanExpression, _ComparisonExpression,
 )
 
 
 def _dupe_ast(ast):
     """
     Create a duplicate of the given AST.
-
-    Note:
-        The comparison expression "leaves", i.e. simple <path> <op> <value>
-        comparisons are currently not duplicated.  I don't think it's necessary
-        as of this writing; they are never changed.  But revisit this if/when
-        necessary.
 
     Args:
         ast: The AST to duplicate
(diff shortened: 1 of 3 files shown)


clenk (approved):
LGTM

PR #544, opened 2022-03-25
Use string comparison in pattern semantic similarity calculations

clenk:
So it will return a % instead of a boolean, to be consistent with the other similarity functions.

### DIFF
--- a/stix2/equivalence/object/__init__.py
+++ b/stix2/equivalence/object/__init__.py
@@ -6,7 +6,7 @@ import time
 
 from ...datastore import DataSource, DataStoreMixin, Filter
 from ...utils import STIXdatetime, parse_into_datetime
-from ..pattern import equivalent_patterns
+from ..pattern import normalize_pattern
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


chisholm (commented):
I think reducing it to string similarity doesn't really work.  I think pattern similarity isn't necessarily accurately reflected in their string similarity.  E.g. consider patterns:

1. `[a:b='c']`
2. `[a:b='somereallyreallyreallyreallylongthinglonglonglong']`
3. `[x:y < 2 OR z:w IN (1,2)]`

For similarity between 1 and 2 I get ~0.1379
For similarity between 1 and 3 I get ~0.1666

So 3 looks more similar to 1 according to the string similarity algorithm.  But actually, I think 1 and 2 are more similar.  They are structurally identical except for a different constant.  But because it takes so many edits to convert from 1 to 2 due to large difference in constants, they have a large edit distance and therefore the algorithm finds a smaller similarity value.

clenk:
We had discussed offline about switching to a tree edit distance algorithm like Zhang-Shasha. Need to look into something like https://github.com/timtadh/zhang-shasha

PR #545, opened 2022-04-29
Complete pattern visitor

### LINKED ISSUE #538: Pattern visitor missing support for certain symbols/components
Hi, I’m working on a translator of STIX Pattern Expressions and I’m leveraging on the pattern visitor you’ve baked into the framework.
I’ve noticed though that some of the Pattern expression symbols/components are not emitted via the instantiate function, that optionally emits custom classes.
For now I had to resort to manually editing patterns.py in multiple parts, but it would be great to have already baked in the support for custom subclasses in all elements. 
This is the list of components where the instantiate function call is missing:
- FollowedByObservationExpression
- StartStopQualifier
- WithinQualifier
- RepeatQualifier
- IntegerConstant
- FloatConstant
- HexConstant
- BinaryConstant
- StringConstant
- BooleanConstant
- TimestampConstant

I don’t know if there’s a reason for why those were left out, guess only because they weren’t needed for the slider.
Anyway the change to make would be, taking for instance the FollowedByObservationExpression, this:
Old:
`return FollowedByObservationExpression([children[0], children[2]])`
New:
`return self.instantiate("FollowedByObservationExpression", [children[0], children[2]])`
I will make a PR that addresses this so to have code references.

chisholm:
Fixes #538 .  Also completes work begun in PR #539 .

Hopefully all AST node classes are overridable now.  Added a few unit tests.

This also addresses an issue where the `module_suffix` and `module_name` parameters to `create_pattern_object()` would be honored on first call and thereafter ignored.  This happened because the class overrides were read and cached in a class-level field, so it was shared among all instances of the visitor.  That means new visitor instances were using overrides set up for old instances.  Unit tests uncovered this problem because the v20 and v21 unit tests use different AST node class overrides.  Whoever ran first would set the class overrides and break the unit tests for whoever ran second.

The new implementation does not cache classes at all.  It loads the module with the classes, and looks up classes directly on that module without caching anything.  Each visitor instance can have a different module instance, so they each can have their own overrides.

### DIFF
--- a/stix2/pattern_visitor.py
+++ b/stix2/pattern_visitor.py
@@ -53,22 +53,32 @@ def same_boolean_operator(current_op, op_token):
 
 
 class STIXPatternVisitorForSTIX2():
-    classes = {}
+
+    def __init__(self, parser_class, module_suffix=None, module_name=None):
+        self.parser_class = parser_class
+
+        if module_suffix and module_name:
+            self.module_suffix = module_suffix
+            self.module = importlib.import_module(module_name)
+        else:
+            self.module_suffix = self.module = None
 
     def get_class(self, class_name):
-        if class_name in STIXPatternVisitorForSTIX2.classes:
-            return STIXPatternVisitorForSTIX2.classes[class_name]
-        else:
-            return None
+        klass = None
+        if self.module:
+            class_name_suffix = class_name + "For" + self.module_suffix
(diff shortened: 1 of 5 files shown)


rpiazza:
@chisholm this looks good to me - much improved!

clenk (approved):
LGTM

PR #555, opened 2022-08-24
Create codeql-analysis.yml

ejratl:
Proposing to enable CodeQL analysis of the code in this repo. The config is auto generated and seems sufficient. The only question is whether we want to enable it or not.


ejratl:
The autogenerated codeql yaml file fails the trailing spaces check in the pre-commit hook which is only called in the 3.9 test and causes the failure. I am looking into how to opt it out of the check.

Issue #569, opened 2023-06-29
Fail to parse valid TLP2.0 marking object.


shellcromancer:
When loading a valid [TLP2.0](https://www.cisa.gov/sites/default/files/2023-02/tlp-2-0-user-guide_508c.pdf) `marking-definition` object such as this, the library will error in `stix2.exceptions.TLPMarkingDefinitionError`

```json
{
            "type": "marking-definition",
            "spec_version": "2.1",
            "id": "marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9",
            "created": "2017-01-20T00:00:00.000Z",
            "definition_type": "tlp",
            "name": "TLP:CLEAR",
            "definition": {
                "tlp": "clear"
            }
}
```

```python
from stix2 import MemoryStore
import requests

stix_json = requests.get('https://github.com/elastic/labs-releases/raw/main/indicators/rustbucket/stix-bundle.json').json()

src = MemoryStore(stix_data=stix_json["objects"])
```

```bash
$ python ingest.py
Traceback (most recent call last):
  File "/Users/foo/Code/ingest.py", line 8, in <module>
    src = MemoryStore(stix_data=stix_json["objects"])
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/datastore/memory.py", line 119, in __init__
    _add(self, stix_data, allow_custom, version)
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/datastore/memory.py", line 35, in _add
    _add(store, stix_obj, allow_custom, version)
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/datastore/memory.py", line 47, in _add
    stix_obj = parse(stix_data, allow_custom, version)
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/parsing.py", line 40, in parse
    obj = dict_to_stix2(obj, allow_custom, version)
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/parsing.py", line 99, in dict_to_stix2
    return obj_class(allow_custom=allow_custom, **stix_dict)
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/v21/common.py", line 226, in __init__
    super(MarkingDefinition, self).__init__(**kwargs)
  File "/Users/foo/Library/Python/3.9/lib/python/site-packages/stix2/base.py", line 232, in __init__
    self._check_object_constraints()
[...]

PR #571, opened 2023-08-17
Adding TLP CLEAR

peasead:
TLP:CLEAR was added as part of TLP 2.0. 

Extension definition:
https://github.com/oasis-open/cti-stix-common-objects/blob/main/objects/extension-definition/extension-definition--60a3c5c5-0d10-413e-aab3-9e08dde9e88d.json

TLP CLEAR marking definition:
https://github.com/oasis-open/cti-stix-common-objects/blob/main/objects/marking-definition/marking-definition--94868c89-83c2-464b-929b-a1a8aa3c8487.json

Resolves https://github.com/oasis-open/cti-python-stix2/issues/569

### DIFF
--- a/stix2/v21/common.py
+++ b/stix2/v21/common.py
@@ -281,6 +281,14 @@ TLP_WHITE = MarkingDefinition(
     definition=TLPMarking(tlp='white'),
 )
 
+TLP_CLEAR = MarkingDefinition(
+    id='marking-definition--94868c89-83c2-464b-929b-a1a8aa3c8487',
+    created='2022-10-01T00:00:00.000Z',
+    definition_type='tlp',
+    name='TLP:CLEAR',
+    definition=TLPMarking(tlp='clear'),
+)
+
 TLP_GREEN = MarkingDefinition(
     id='marking-definition--34098fce-860f-48ae-8e50-ebd3cc5e41da',
     created='2017-01-20T00:00:00.000Z',


rpiazza:
Hi @peasead,

I don't think this is the correct way to add TLP 2.0 CLEAR.  The object exists in the common object repository, so you should use those.  Remember, any STIX object of the same id and same version must be identical.  Also, what you have done will probably not validate using the extension definition's schema or the base schema - since there is no TLP:CLEAR in TLP 1.0.

I suggest you add all five TLP 2.0 data marking objects to your environment, and then do something locally like you did in this pull request for all five.

peasead:
Thanks @rpiazza. I've not worked with the project beyond using it, so sorry if this wasn't correct.

> I suggest you add all five TLP 2.0 data marking objects to your environment, and then do something locally like you did in this pull request for all five.

For the errors [here](https://github.com/oasis-open/cti-python-stix2/issues/569), this has to be fixed locally - only in my environment, not for the whole project? 

> Also, what you have done will probably not validate using the extension definition's schema or the base schema - since there is no TLP:CLEAR in TLP 1.0.

I assumed the path `stix2/v21` would mean that this would only be for TLP 2, but maybe that's not correct?

Thanks again for the help.

chisholm:
> > Also, what you have done will probably not validate using the extension definition's schema or the base schema - since there is no TLP:CLEAR in TLP 1.0.
> 
> I assumed the path `stix2/v21` would mean that this would only be for TLP 2, but maybe that's not correct?
> 
> Thanks again for the help.

Extension definitions reference schemas via the required `schema` property.  The TLP 2 extension definition references [https://github.com/oasis-open/cti-stix-common-objects/tree/master/extension-definition-specifications/tlp-2.0](https://github.com/oasis-open/cti-stix-common-objects/tree/master/extension-definition-specifications/tlp-2.0), which resolves to a directory; the schema is a [file](https://github.com/oasis-open/cti-stix-common-objects/blob/main/extension-definition-specifications/tlp-2.0/extension-definition--60a3c5c5-0d10-413e-aab3-9e08dde9e88d.json) in that directory.  TLP 2 markings which use that extension definition must be valid against that schema.  Your proposed change results in a STIX object which is not schema-valid.  It is also not spec-valid, since the spec says TLP markings (with `definition_type="tlp"`, as your code has) *must* be one of those the spec [defines](https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_yd3ar14ekwrs).  If you wanted to create a marking that way, you would have to use a different definition_type.

But `definition` and `definition_type` are deprecated properties: they should not be used anyway.  The best way to do it is via the officially supported extension mechanism.  That is what the objects in the common object repository do.  If you wanted to add predefined objects to the library (or to your own code) which correspond to the common object repository markings, you'd have to create them with *exactly* that set of properties.

For example:

```python
tlp_20_clear = stix2.MarkingDefinition(
    id="marking-definition--94868c89-83c2-464b-929b-a1a8aa3c8487",
    created="2022-10-01T00:00:00.000Z",
    name="TLP:CLEAR",
    extensions={
        "extension-definition--60a3c5c5-0d10-413e-aab3-9e08dde9e88d": {
            "extension_type": stix2.v21.vocab.EXTENSION_TYPE_PROPERTY_EXTENSION,
            "tlp_2_0": "clear"
        }
    }
)
```

One might also consider registering an extension object (e.g. with `@CustomExtension`), but you'd have to ensure somehow that the `tlp_2_0` property in the extension matches with the `name` top-level property, so people couldn't accidentally create malformed markings.  The above is just a simple demo for illustration.

PR #575, opened 2023-10-18
Fix properties types

SYNchroACK:
- Rename properties types based on the naming pattern
- Add missing _type field

### DIFF
--- a/stix2/v21/observables.py
+++ b/stix2/v21/observables.py
@@ -143,11 +143,12 @@ class EmailAddress(_Observable):
     _id_contributing_properties = ["value"]
 
 
-class EmailMIMEComponent(_STIXBase21):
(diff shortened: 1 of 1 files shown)


rpiazza:
@chisholm - is this worth merging?

chisholm:
In principle, I am okay with what the changes are doing.  But this doesn't appear to change type names in enough places.  For example, see imports in:

https://github.com/oasis-open/cti-python-stix2/blob/6dd83fae9dfa6b3fda9de865801d5796f9d8f7e7/stix2/v21/__init__.py#L26-L35

This change breaks some of those imports.  Perhaps an IDE should be used to do the refactor, to ensure the names are changed in *all* places they are used.

Issue #577, opened 2023-11-13
Objects containing LIstProperty or DictionaryProperty fields are not Immutable


robthew:
Although STIX objects are supposed to be immutable, the contents of List or Dictionary properties can be modified using operations line 'clear' and 'append'. Simple types like Strings are immutable, and you can't replace a List or Dictionary - this would change the object's ID, triggering an immutability error - but you can modify List or Dictionary contents.

For instance, if you have an Identity object:
```
toplevel_org = Identity(
    x_comment="#1.b. Organization",
    name="Organization1",
    description="Test org.",
    identity_class="critical-infrastructure-or-private",
    roles=["director"],
    allow_custom=True,
)
```
This would throw an error because it changes the object ID of 'roles':
```
toplevel_org.roles=["cleared-dod-contractor"]
```

But, you can accomplish the same thing with 'clear' and 'append':
```
toplevel_org.roles.clear()
toplevel_org.roles.append("cleared-dod-contractor")
```

There does not seem to be a simple fix to this, but it's something to be aware of.

chisholm:
Concur.  Any mutable property value gives an opening for a user to change the object.  The STIX spec says all STIX objects are immutable; the stix2 library wards off the more casual/simple attempts to change an object.  But it does not enforce deep immutability, so it is possible to change the object in sneakier ways.

I was not involved with the original design, but maybe this never rose to the level of me pushing for changes, because I liked the simplicity and it seemed unlikely to cause problems.  If we change property types (e.g. change list valued properties to immutable tuple-valued properties), it would be backward-incompatible with respect to type-sensitive code like `isinstance(obj.labels, list)` (to pick a common list-valued property).  Not sure how important that would be.

PR #580, opened 2023-11-29
Added documentation about immutability issues

robthew:


robthew:
This addresses issue #577, adding documentation to the overview and an example to the creation notebook.

PR #587, opened 2024-03-16
Upgrade python

rpiazza:
@clenk 

Can you look into this?   For some reason the pre-commit is not working.  Nothing stands out - see the actions


clenk:
Think it might be caused by https://github.com/PyCQA/isort/issues/2077 so we should upgrade the version of isort used in https://github.com/oasis-open/cti-python-stix2/blob/master/.pre-commit-config.yaml#L19.

rpiazza:
I guess I need to do it because of permissions, right?

PR #591, opened 2024-04-04
Adding enumproperty to simple type and edit openvocabproperty

ryantxu1:


rpiazza:
- I'm not sure you need to have that new test in the enum clean if you pass strict to the super class (StringProperty)
- Can you use strict in OpenVocabProperty and uncomment out the code that checks for allowed when strict is true?

rpiazza:
Looks good, but can you add to the comment saying that we are doing this new "strict" thing.

@chisholm - can you do a quick review - since you know the clean stuff so well

chisholm:
These comments are not just about this PR I think, but other related introduced code.

https://github.com/oasis-open/cti-python-stix2/blob/dce689a6a05e163cf9908caf4311db2d99286fbe/stix2/properties.py#L425

What is "this value" in the error message?  If there was a value, it would probably be helpful to include it in the error message.  If there's something wrong with a value, you should show the value so people can see what it is that's wrong.

https://github.com/oasis-open/cti-python-stix2/blob/dce689a6a05e163cf9908caf4311db2d99286fbe/stix2/properties.py#L457
https://github.com/oasis-open/cti-python-stix2/blob/dce689a6a05e163cf9908caf4311db2d99286fbe/stix2/properties.py#L462

In one place `strict` is used, in the other it's `strict_flag`.

Also, `allow_custom=False` is hardcoded in both calls.  I think that's incorrect: the allow_custom flag passed into the .clean() method should propagate.  If customization is allowed, it is allowed *throughout* the whole object.  I don't actually think the allow_custom flag is used in the "simple" property class .clean() methods, but in principle it should propagate.

https://github.com/ryantxu1/cti-python-stix2/blob/917a4cc449bf0aa1589363ae4d8425ed9291bcda/stix2/properties.py#L752-L755

Should remove the "disabled" comment if the code is no longer disabled.

A stylistic comment:

https://github.com/ryantxu1/cti-python-stix2/blob/917a4cc449bf0aa1589363ae4d8425ed9291bcda/stix2/properties.py#L756

I don't think it's ever necessary/useful to compare a boolean variable to a boolean literal like that.  We write comparison expressions to obtain something which evaluates to a boolean (or something truthy/falsey, depending on the language).  If you have a boolean variable and just want to check it, you have what you need already.  E.g.

```python
if strict:
    pass

if not strict:
    pass

if strict or not something_else:
    pass
```

but not:

```python
if strict is True:
    pass

if strict == False:
    pass

if strict is True or something_else is False:
    pass
```
etc.

chisholm:
Do you mean it's not propagated from Dictionary.clean() to the other .clean() methods it calls?  We need to be able to discriminate the various types of values inside the mapping, which is done here by attempting a clean() call.  Without strict=True, StringProperty.clean() would never fail and everything would be converted into a string, for example.  That could corrupt a user's dictionary value.

One could consider what a strict dictionary clean() ought to be, if anything.  We could say that in strict mode, only dicts (and/or maybe Mappings) are accepted.  As it stands, if a string is passed, it will attempt to interpret as JSON and parse it (see _get_dict()).  That doesn't seem very "strict".

rpiazza:
What is was referring to was that I expected strict would be a keyword argument to all clean methods, but it isn't for Dictionary:

https://github.com/ryantxu1/cti-python-stix2/blob/relational-data-sink/stix2/properties.py#L431

Doesn't whether or not you want to be strict come from the initial call to clean?

chisholm:
> What is was referring to was that I expected strict would be a keyword argument to all clean methods, but it isn't for Dictionary:

Yes!  I made that point in another issue/PR... can't find it now.  All overrides of Property.clean() really should be valid, and right now they are not.  My IDE notices this too and shows the squiggly underlines.  There is hackage in ListProperty to deal with inconsistent signatures:

https://github.com/ryantxu1/cti-python-stix2/blob/917a4cc449bf0aa1589363ae4d8425ed9291bcda/stix2/properties.py#L244-L245

which is ugly and I don't like it and I had wanted it to be removed... but on further reflection, it's probably necessary in case users have written custom Property subclasses whose .clean() methods use the old signature.  But *our* .clean() methods should have valid signatures.

> Doesn't whether or not you want to be strict come from the initial call to clean?

It presently does not:

https://github.com/oasis-open/cti-python-stix2/blob/74873aadc0385b84e126407be6be933815d4bd20/stix2/base.py#L50-L52

I guess it would be possible to support a "strict" keyword arg when creating a STIX object instance, similar to the allow_custom setting we have now.  It could in theory be passed in to the top-level .clean() call, but right now that's not supported.  So at the top-level, strict will always have its default value (False), which is backward-compatible with the original lenient cleaning behavior.  It's only true in the particular case where we need it to be, which is this "union" style property type enforcement in DictionaryProperty.  A union property type doesn't occur anywhere else for now.

In fact, I wonder if we could/should have just created a new general-purpose "UnionProperty" class, instead of creating a special case inside DictionaryProperty... Who knows, some future STIX object might define a property that way (e.g. value could be int or string), and UnionProperty would be useful in that case.

PR #592, opened 2024-04-05
Add required ReadTheDocs configuration

clenk:
Should fix broken RTD doc builds. Needed because of https://blog.readthedocs.com/migrate-configuration-v2/

@rpiazza


adulau:
It makes sense.

PR #605, opened 2024-09-23
Fix constraint violation checking for GranularMarking in v21.

### LINKED ISSUE #586: Granular marking constraint checking is incorrect
The STIX 2.1 spec [section 7.2.3](https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_robezi5egfdr) says that one cannot include both `lang` and `marking_ref` in a granular marking:

> Because each granular marking instance applies to either a language or a marking, one and only one of these properties MUST be present on each instance of a granular marking.

The library does not check this:

```python
import stix2

granular_marking = stix2.GranularMarking(
  lang="en",
  marking_ref=stix2.TLP_GREEN,
  selectors=["foo"]
)

granular_marking.fp_serialize(sys.stdout, indent=4)
```

results in:

```json
{
    "lang": "en",
    "marking_ref": "marking-definition--34098fce-860f-48ae-8e50-ebd3cc5e41da",
    "selectors": [
        "foo"
    ]
}
```

which is invalid.

mfang90739:
This PR fixes #586.

- used the `_check_mutually_exclusive_properties` method to catch the exception caused by the incorrect use of both `lang` and `marking_ref` properties in the same time
- added a corresponding unit test case

### DIFF
--- a/stix2/test/v21/test_markings.py
+++ b/stix2/test/v21/test_markings.py
@@ -175,7 +175,7 @@ def test_granular_marking_mutual_exclusion_error():
         stix2.v21.GranularMarking(
             lang="en",
             marking_ref=stix2.TLP_GREEN,
-            selectors=["foo"]
+            selectors=["foo"],
         )
     assert excinfo.value.cls == stix2.v21.GranularMarking
     assert excinfo.value.properties == ["lang", "marking_ref"]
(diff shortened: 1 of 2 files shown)


rpiazza:
@mfang90739 -
The linting/style checking failed (see 3.11 build).  Just some minor issues.  Can you fix them? Otherwise looks good!

rpiazza:
One last thing to fix to make pre-commit happy...

Check project styling....................................................Failed
  - hook id: flake8
  - exit code: 1
  
  stix2/v21/common.py:82:1: E302 expected 2 blank lines, found 1

Issue #572, opened 2023-09-07
Location Object can't have 0 for Longitude/Latitude


nyc3-djigarjian:
When attempting to create Location objects with '0' as a value for either the latitude or longitude, the library raises a DependentPropertiesError when it should not. See relevant portion of the StackTrace below:  

`country = stix2.Location( File "/usr/local/lib/python3.9/site-packages/stix2/base.py", line 232, in __init__ self._check_object_constraints() File "/usr/local/lib/python3.9/site-packages/stix2/v21/sdo.py", line 398, in _check_object_constraints self._check_properties_dependency(['longitude'], ['latitude']) File "/usr/local/lib/python3.9/site-packages/stix2/base.py", line 109, in _check_properties_dependency raise DependentPropertiesError(self.__class__, failed_dependency_pairs) stix2.exceptions.DependentPropertiesError: The property dependencies for Location: (longitude, latitude) are not met.`

I've traced the issue to the stix2\base.py _check_properties_dependency method where the function uses the "not" operator to check if a property is a valid value. In this scenario, '0' is a valid Longitude and Latitude and an additional layer of checking needs to occur for Number values. **Note: Issue arises at ```if not self.get(p)```**

```
    def _check_properties_dependency(self, list_of_properties, list_of_dependent_properties):
        failed_dependency_pairs = []
        for p in list_of_properties:
            for dp in list_of_dependent_properties:
                if not self.get(p) and self.get(dp):
                    failed_dependency_pairs.append((p, dp))
        if failed_dependency_pairs:
            raise DependentPropertiesError(self.__class__, failed_dependency_pairs)
```

ejratl:
Thanks for the bug report and for tracking down the exact location in the code where this occurs! TIL: 0°N 0°E is known as "Null Island".

rpiazza:
_check_properties_dependency mentioned above can be found at 

https://github.com/oasis-open/cti-python-stix2/blob/master/stix2/base.py#L102 

I think the problem is that the get method doesn't take into consideration that a value of 0 will be interpreted as None.  If the property has a value of 0, then it exists, and shouldn't cause this error.

PR #607, opened 2024-10-17
Fix for issue #572

HackerShark:
Issue #572 errors when longitude or latitude has a 0 value. This fix takes into account 0 values for longitude and latitude.

### DIFF
--- a/stix2/base.py
+++ b/stix2/base.py
@@ -103,7 +103,8 @@ class _STIXBase(collections.abc.Mapping):
         failed_dependency_pairs = []
         for p in list_of_properties:
             for dp in list_of_dependent_properties:
-                if not self.get(p) and self.get(dp):
+                if self.get(p) is None and self.get(dp) is not None:
+                # if not self.get(p) and self.get(dp):
                     failed_dependency_pairs.append((p, dp))
         if failed_dependency_pairs:
             raise DependentPropertiesError(self.__class__, failed_dependency_pairs)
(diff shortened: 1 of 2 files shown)


rpiazza:
Seems line two tests failed this time.  Before you push again, run the tests in pycharm.  Just click right on test-location.py and select test

rpiazza:
Some other test, not in test-location.py, failed .  Can look into that?

rpiazza:
It seems to be failing in the v20 tests

Any progress?

HackerShark:
Tested the code locally and it passed. This fix should be correct.

rpiazza:
Some minor style issues.  If you fix them I can merge :-)

chisholm:
This is still incorrect.  A boolean property can be `False` and present:

```python
class FalseyTest(_STIXBase):
    _type = "falsey"
    _properties = {
        "type": TypeProperty(_type, spec_version='2.1'),
        "id": IDProperty(_type, spec_version='2.1'),
        "foo": StringProperty(),
        "falsey1": BooleanProperty(),
    }

    def _check_object_constraints(self):
        super()._check_object_constraints()
        super()._check_properties_dependency(["falsey1"], ["foo"])

# Should not throw since both props are present.
FalseyTest(foo="bar", falsey1=False)
```

still yields:

```
stix2.exceptions.DependentPropertiesError: The property dependencies for FalseyTest: (falsey1, foo) are not met.
```

The problem is that a falsey test was done in a context where it did not correctly substitute for a presence test.  Sometimes you can get away with that, sometimes you can't.  A direct presence test is better here, e.g. `if p in self...`.

rpiazza:
@HackerShark - still some linty errors - see tox run.

@chisholm - when Mo fixed these errors, and you review his change for me?

PR #608, opened 2024-11-06
Experimental changes to allow relational-data-sink implementation to work with different database backends

mfang90739:
Added test code to scope out the work needed to support the following variants:

- SQLite
- MariaDB
- MySQL
- Microsoft SQL (MSSQL)

Testing has been done on a MacBook Pro M3 system for the first three; the corresponding log file is attached for reference. The code for MSSQL remains to be tested, so consider it as a place holder for now.

[sqlite.log](https://github.com/user-attachments/files/17652847/sqlite.log)
[mariadb.log](https://github.com/user-attachments/files/17652850/mariadb.log)
[mysql.log](https://github.com/user-attachments/files/17652853/mysql.log)

### DIFF
--- a/stix2/datastore/relational_db/relational_db.py
+++ b/stix2/datastore/relational_db/relational_db.py
@@ -80,7 +80,41 @@ class RelationalDBStore(DataStoreMixin):
                 auto-detect all classes and create table schemas for all of
                 them.
         """
+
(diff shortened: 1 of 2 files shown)


mfang90739:
Just pushed back the change for MS-SQL. The testing result is captured in 
[mssql.log](https://github.com/user-attachments/files/17709122/mssql.log)

PR #612, opened 2024-11-21
handle-hex-better

rpiazza:
if hex is being stored in the db as a string - don't convert it to bytes

### DIFF
--- a/stix2/datastore/relational_db/database_backends/database_backend_base.py
+++ b/stix2/datastore/relational_db/database_backends/database_backend_base.py
@@ -105,3 +105,10 @@ class DatabaseBackend:
     @staticmethod
     def array_allowed():
         return False
+
+    def generate_value(self, stix_type, value):
+        sql_type = stix_type.determine_sql_type(self)
+        if sql_type == self.determine_sql_type_for_string_property():
+            return value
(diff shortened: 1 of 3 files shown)


chisholm:
`generate_value()` has a very general name, but seems to do a very specific thing.  In words, if backend says "use string type", `value` is returned; if backend says "use hex type" and that type is not the same as the string type, assume `value` is hex, decode that to bytes, and return it.  The method name should be a lot more specific, I think.

It would seem to let backend authors choose a non-string SQL type for hex properties if they wish and have the datastore handle it gracefully.  But currently if they choose anything other than string, it crashes, and there's no way to fix that.  If not a string, if they choose a non-binary type, it doesn't make any sense (you'd be inserting a binary value into a non-binary column).

It seems like an idiosyncratic bit of flexibility which no one can use.  Why was this attention afforded to hex properties but not binary?  Wouldn't you want to be able to store STIX binary values in binary columns too?

I think we should just hard code it to text for now for hex and binary.  Flexibility can be added later if it can be made to work.

rpiazza:
Yes - I was thinking of it as a general method for all special cases, for which now there is only hex.  
I was thinking it could be overridden for a particular database.  It only doesn't work for dictionaries...

PR #615, opened 2024-12-13
Add support of TLP:AMBER+STRICT (#583)

### LINKED ISSUE #583: CISA TLP definition updates
CISA has new guidance on TLP markings and modifications... the stix2(.1) common.py configs do not reflect this more recent guidance...

[https://www.cisa.gov/news-events/news/traffic-light-protocol-tlp-definitions-and-usage](https://www.cisa.gov/news-events/news/traffic-light-protocol-tlp-definitions-and-usage)

TLP-WHITE --> TLP-CLEAR
+TLP-AMBER+STRICT

These markings should be included in common config as they're officially part of the guidance now and already in use and will be in use going forward.

Kedae:
as defined here : https://www.cisa.gov/news-events/news/traffic-light-protocol-tlp-definitions-and-usage

Closes #583

### DIFF
--- a/stix2/v21/common.py
+++ b/stix2/v21/common.py
@@ -298,6 +298,14 @@ TLP_AMBER = MarkingDefinition(
     definition=TLPMarking(tlp='amber'),
 )
 
+TLP_AMBER_STRICT = MarkingDefinition(
+    id='marking-definition--939a9414-2ddd-4d32-a0cd-375ea402b003',
+    created='2022-08-16T00:00:00.000Z',
+    definition_type='tlp',
+    name='TLP:AMBER+STRICT',
+    definition=TLPMarking(tlp='amber+strict'),
+)
+
 TLP_RED = MarkingDefinition(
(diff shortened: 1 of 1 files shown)


chisholm:
Presumably you've read the linked issue 583 which refers to the common object repository.  This implementation does not match the marking definition defined [there](https://github.com/oasis-open/cti-stix-common-objects/blob/main/extension-definition-specifications/tlp-2.0/examples/tlp-amber%2Bstrict.json).  I am not a maintainer, but if you wanted to contribute a tlp 2.0 implementation, seems like you'd need it to match those objects.  Also, I'd think you'd want to implement all tlp2 markings, not just one of them.

rpiazza:
We cannot just change python stix2 to include something not defined in the 2.1 specification document.  If you want to use STIX objects not defined in the 2.1 specification, you need to use an extension, as defined in section 7.3 of the 2.1 specification.  The OASIS CTI TC has provided the one in the common object repository - https://github.com/oasis-open/cti-stix-common-objects/tree/main/extension-definition-specifications/tlp-2.0.

PR #621, opened 2025-03-09
Fix integer constraint

rpiazza:
Can you review this and merge it?

I wasn't creating the integer min/max constraint in the db table


adulau:
Adding the `min`and `max` constraint makes sense. Thanks for the PR. 

Not exactly sure why the test is failing.

chisholm:
On postgres: I do see the check constraints in the table schemas, with the correct limits, so the constraint creation seems to work.  Unit tests pass locally, without tox, with the usual unexpected connection drop in the last test (still need to figure that one out).  So it works no worse than before. :)

I just tried the sqlite backend for the first time, and the query side failed.  It was still trying to qualify table names with schema names.  It's not too surprising it didn't work since I don't think the other backends existed when I was writing it.  So that will need more work.  It probably won't work with Mariadb either.

I don't have permission to merge anything though.

rpiazza:
@adulau - Andy and I have been working on a relational database DataStore for Python STIX.  I will be doing the merge to the relational-data-sink branch once @chisholm fixes the tests.  

I'm not sure if we will be merging the relational-data-sink branch into the master branch at this time

chisholm:
Yep, all sqlite tests pass except that one.  Only one more to go!

I tried running all sqlite and postgresql tests together, and encountered new errors.  I tracked it down to a simple contextmanager cleanup error I made.  I fixed it in PR #630 .  With that fix, I was able to get a complete run of all relational datastore unit tests so far, to work, except for that one sqlite test with the email message.

Next, getting mariadb to work! :)

chisholm:
Maybe should continue the discussion here instead of a closed PR:

> The some rdb tests still fail not just the email one

When running just the rdb tests, I get:

    FAILED stix2/test/v21/test_datastore_relational_db.py::test_multipart_email_msg[sqlite] - AssertionError: assert {'Content-Dis...98.51.100.3']} == {'X-Originati...98.51.100.3']}
    ===== 1 failed, 371 passed, 108 xfailed, 32 warnings in 117.33s (0:01:57) =====

So it is just the email test for me.  When running all unit tests, I get a lot more errors, including a lot with the relational datastore unit tests.  That suggests bad interactions with other unit tests.  So the obvious thing to try is to fix other unit tests to see if the relational datastore unit test errors will go away.

Were you thinking we should try to fix the other unit tests (or whatever they're testing) too?

Issue #578, opened 2023-11-19
Cannot accept timestamps more precise than microseconds because of datetime module limitation


zed-eiq:
👋 hello folks! While implementing support for an intel vendor's API, our application could not process their STIX 2.1 objects.

Turns out that the vendor is sending us timestamps with nanosecond precision (e.g.: `2023-11-12T14:16:48.1699798608Z`), which then causes `datetime.strptime` to fail when we attempt to parse with `stix2`.

We checked the STIX 2.1 specification, and it only discourages timestamps more precise than nanoseconds, so the vendor is in this sense spec-compliant: https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_i7rrdtzgta8d

Relevant traceback:

```
File "/usr/lib/python3.8/_strptime.py", line 349, in _strptime\n    raise ValueError("time data %r does not match format %r" %\nValueError: time data \'2023-11-12T14:16:48.1699798608Z\' does not match format \'%Y-%m-%dT%H:%M:%S.%fZ\'\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  
File "<REDACTED>/env/lib/python3.8/site-packages/stix2/base.py", line 50, in _check_property\n    kwargs[prop_name], has_custom = prop.clean(\n  
File "<REDACTED>/env/lib/python3.8/site-packages/stix2/properties.py", line 383, in clean\n    return parse_into_datetime(\n  
File "<REDACTED>/env/lib/python3.8/site-packages/stix2/utils.py", line 252, in parse_into_datetime\n    raise ValueError(\nValueError: must be a datetime object, date object, or timestamp string in a recognizable format.\n\nThe above exception was the direct cause of the following exception:\n\nTraceback (most recent call last):\n  
```

We've implemented a workaround to round all timestamps to the nearest microsecond. But this mutates the original data, so it's no longer strictly the same STIX object.

What would the OASIS Open TC org recommend?

-----
Full traceback for context/transparency:

```
WARNING  eiq.extensions.stix_taxii2.protocol:protocol.py:195 {'id': 'indicator--7032a43b-4263-4a26-935c-8aed935dc172', 'event': 'Skipping invalid STIX 2.1 object', 'logger': 'eiq.extensions.stix_taxii2.protocol', 'level': 'warning', 'timestamp': '2023-11-13T12:22:26.052552Z', 'exception': 'Traceback (most recent call last):\n  
[...]

PR #624, opened 2025-03-17
adjust datetime str nanoseconds to microseconds

ankit-eiq:
This pull request addresses the issue described in [#578](https://github.com/oasis-open/cti-python-stix2/issues/578), where handling of timestamp precision was not correctly adjusted when converting nanoseconds to microseconds.

**Changes made:**
Implemented logic to adjust nanoseconds to microseconds precision when the timestamp has nanoseconds precision (9 digits followed by 'Z').

**Testing:**
Updated unit tests to check for timestamp conversion and precision handling with different values (e.g., second, millisecond, nanosecond).
Validated that timestamp precision is correctly truncated for nanosecond precision when required.

### DIFF
--- a/stix2/test/v21/test_timestamp_precision.py
+++ b/stix2/test/v21/test_timestamp_precision.py
@@ -71,6 +71,11 @@ def test_stix_datetime():
 
 @pytest.mark.parametrize(
     "us, precision, precision_constraint, expected_truncated_us", [
+        (123456789, Precision.ANY, PrecisionConstraint.EXACT, 123456),
+        (123456789, Precision.SECOND, PrecisionConstraint.EXACT, 0),
+        (123456789, Precision.SECOND, PrecisionConstraint.MIN, 123456),
+        (123456789, Precision.MILLISECOND, PrecisionConstraint.EXACT, 123000),
+        (123456789, Precision.MILLISECOND, PrecisionConstraint.MIN, 123456),
         (123456, Precision.ANY, PrecisionConstraint.EXACT, 123456),
         (123456, Precision.SECOND, PrecisionConstraint.EXACT, 0),
         (123456, Precision.SECOND, PrecisionConstraint.MIN, 123456),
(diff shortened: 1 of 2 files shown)


rpiazza:
@chisholm - can you look at this to see if it is worth being merged?

chisholm:
This PR does not address the larger issue, which is that all implementations have limits, and those limits may or may not satisfy a particular user's requirements.  If an implementation's limits are below what a user requires, one can try to extend the limits a little farther, but there are still limits.  The next user might need the limit extended again, and again... this is impossible to "fix".

I didn't read that issue as claiming the library has a problem that needs a fix.  It pointed out a limitation, and asked how should implementations deal with situations like that.  They apparently implemented a workaround which worked for them (unclear if it entailed forking/changing the library at all), but it may not work for all.  They could afford to throw away some precision, but others may not be willing/able to do that.  They were really asking what should an implementation do when asked to process data which exceeds the implementation's limits.  I don't think there's any one right answer to that.  The conservative approach is to error out, which is what this library does.  That ensures that the user is aware of those limits and doesn't get any nasty surprises.  Silently changing data could be considered more dangerous, and if that data is from an established STIX object, it implies we could be changing the object, which is a spec compliance violation.

So perhaps the best thing to do is leave the aforementioned conservative approach as this implementation, and let library users address the limitations in their own individual ways, rather than subjecting all users to a seemingly one-size-fits-all approach which may not actually work for everyone.

If I were to consider this PR, observations include:
- This implementation truncates extra digits; the issue specifically refers to "rounding", so this implementation may not be exactly equivalent to their approach.
- It errors on 7, 8 fractional second digits, succeeds with 9 digits, errors with 10.  Why so specific to exactly 9 digits?  If you're going to implement via string operations, you could handle as many digits as you wanted.  Why not generalize?
- If nanosecond precision is so common that we really need to be able handle it losslessly somehow, I guess that means we'd need to try to find another Python library which can do that, and stop using builtin datetime.

PR #630, opened 2025-04-10
Fix unit test cleanup bug in relational datastore test suite

chisholm:
This fixes my goof where I used a try-except where I should have used a try-finally for a unit test cleanup.  This caused some unit tests to not be properly cleaned up unless an exception occurred, which could cause subsequent unit test failures.


rpiazza:
The some rdb tests still fail  not just the email one

PR #635, opened 2025-04-24
fix email message test

rpiazza:


chisholm:
I think the real fix is in input_creation.py; the rest is stylistic reformatting.  Tests pass locally too, looks good to me!

